PR #12, opened 2023-04-18
Set up open api

alvaro:
This PR include changes to add:
* Minimal open api configuration
* Open api generate plugin configuration
* A test endpoint


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/HmppsAccreditedProgrammesApi.kt
```diff
@@ -1,11 +0,0 @@
-package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi
-
-import org.springframework.boot.autoconfigure.SpringBootApplication
-import org.springframework.boot.runApplication
-
-@SpringBootApplication()
-class HmppsAccreditedProgrammesApi
-
-fun main(args: Array<String>) {
```
Gweaton, line 9: Ah, so this is why it was so hard to find the `main` method in the example project we looked at?

Gweaton (approved):
This looks good to me - thanks for sorting. It's great to have something we can test against with the UI to check everything is configured correctly!

For future PRs, it would be great if we could squash and delete unused / changed commits that don't need to live on `main` - e.g. ideally we could delete the commit for adding the `HelloController` once the new `ProgrammesController` was added. This helps to keep the `main` branch clear of commits that aren't actually used, and means there's less code for a reviewer to review.

So for this PR, the commits would look a bit like:
- Add initial open api and api generator plugin
- Remove class including main method...
- Add /programmes endpoint including test data
- Fix spelling (this would fix spelling introduced in a previous branch, but any mistakes introduced in this branch would be squashed into the commit they were introduced in)
- Updated README with instructions to access the open api documentation

No need to worry about this for this PR, just something to consider in future

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/HmppsAccreditedProgrammesApi.kt
alvaro: Yes, it is generated by the plugin and if you add another main method then you get an error

PR #14, opened 2023-04-19
(336) Set up ADR directory and document the data structures

Gweaton:
## Context

https://trello.com/c/oIyZQRPb/336-document-the-data-structure-in-the-ui-repo

## Changes in this PR

- Adds a new ADR directory using adr-tools for documenting future API decisions.
- Adds UML diagram of the data structures we'll be providing in our API.

image


doc/adr/0001-record-architecture-decisions.md
```diff
@@ -0,0 +1,19 @@
+# 1. Record architecture decisions
```
yndajas, line 1: Is it an architectural decision to have architectural decision records? 🤯

doc/architecture/diagrams/data-structures/programmes-organisations-people-referrals.wsd
```diff
@@ -0,0 +1,90 @@
+@startuml
```
yndajas, line 1: Is it worth documenting how to use this file? Does it need a StarUML licence?

yndajas (approved):
Looks good. One comment and one question on the files, plus:

Are the data structures really documented as an ADR (per the PR and branch titles)? It looks like this adds one ADR which is to use ADRs (meta), then the second commit is just adding a diagram and a UML file without an accompanying ADR (relating to the particular decisions that have been made)? Or does a diagram without an accompanying numbered Markdown file count as an ADR?

doc/adr/0001-record-architecture-decisions.md
Gweaton: This is automatically generated by [adr-tools](https://github.com/npryce/adr-tools) but yes!

doc/architecture/diagrams/data-structures/programmes-organisations-people-referrals.wsd
Gweaton: It doesn't, we can copy the code into something like PlantUML to tweak it, that's what we used to generate the diagrams. Would documenting this in the commit message be enough, do you think?

yndajas: I think that's probably fine, unless other devs are likely to need to get involved in that, in which case it might be nice to add to `/doc` (but I doubt that's the case)

Gweaton:
@yndajas I suppose the diagram isn't an official ADR (in terms of the ADR tool we use to record these decisions), but it is a record of our architectural decisions! I don't think we need to add an accompanying "official" ADR for this, as we tend to use those for things more like tooling choices or more complex architectural issues than an underlying data structure.

yndajas:
[Document the data structure as an ADR in the API repo](https://trello.com/c/oIyZQRPb/336-document-the-data-structure-as-an-adr-in-the-api-repo)

PR #23, opened 2023-05-18
Implement the GET /courses/{courseId}/offerings/{offeringId} end-point

bell-pa:
## Context

Trello ticket [346]( https://trello.com/c/90RhMAnl/346-add-api-endpoint-for-an-individual-courseoffering)

## Changes in this PR

* New GET `/courses/{courseId}/offerings/{offeringId}` end-point returns information about a single course offering
* Returns 404 NOT_FOUND if the ids do not select an offering
* Added full stack test of new end-point. (Integration test)

Notes:
* hmps-approved-premises-api uses  library ["org.zalando:problem-spring-web-starter:0.27.0"](https://github.com/zalando/problem-spring-web)  to assist with handling HTTP error responses following [RFC 7807](https://datatracker.ietf.org/doc/html/rfc7807).  Currently this application does not do so and does not adhere to the response format defined in rfc7807. Should it?
* The `GET /courses/{courseId}/offerings` returns an empty response when the supplied `courseId` does not select a course.  Should it return 404 NOT FOUND instead?

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/config/HmppsAccreditedProgrammesApiExceptionHandler.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/config/HmppsAccreditedProgrammesApiExceptionHandler.kt
@@ -4,9 +4,11 @@ import org.slf4j.LoggerFactory
 import org.springframework.http.HttpStatus
 import org.springframework.http.HttpStatus.BAD_REQUEST
 import org.springframework.http.HttpStatus.INTERNAL_SERVER_ERROR
+import org.springframework.http.HttpStatus.NOT_FOUND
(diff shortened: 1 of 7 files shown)


Gweaton (approved):
Thanks for this. The approach looks good to me.

On your comments:

> hmps-approved-premises-api uses library ["org.zalando:problem-spring-web-starter:0.27.0"](https://github.com/zalando/problem-spring-web) to assist with handling HTTP error responses following [RFC 7807](https://datatracker.ietf.org/doc/html/rfc7807). Currently this application does not do so and does not adhere to the response format defined in rfc7807. Should it?

Yes - it would be good to make sure we're adhering to that response format, are we missing a `details`-type field on the error response at the moment? Is there anything else we could be including?

> The GET /courses/{courseId}/offerings returns an empty response when the supplied courseId does not select a course. Should it return 404 NOT FOUND instead?

That would make sense to me!

PR #46, opened 2023-06-14
Improve docker-compose.yml configuration.

bell-pa:
## Changes in this PR
Update docker-compose.yml:
* api container waits for postgresql instance to become available before starting.
* Database is empty because this is using a production image.
* Db can be populated from `R_test_data.sql` using intelliJ or the postgres command line tools available in the running container.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
       - APPLICATION_AUTHENTICATION_UI_ALLOWLIST=0.0.0.0/0
       - SPRING_H2_CONSOLE_SETTINGS_WEBALLOWOTHERS=true
 
-  posgresql:
+  postgresql:
     image: postgres
     container_name: postgres
     networks:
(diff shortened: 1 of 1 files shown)


Gweaton (commented):
Thanks for this - looks good to me.

Re: the following, could you provide an example postgres command for how we can run the commands here? Let's add this to the README too so it's obvious how to do this.

> Db can be populated from R_test_data.sql using intelliJ or the postgres command line tools available in the running container.


Could you please add more description to the commit message? There's some context in the PR description that would be really useful to put here, so a useful commit message for future readers looking back at why this file changed would see the following:

```
Improve docker-compose.yml configuration

Run API as part of docker-compose configuration, waiting for Postgresql instance to be ready before starting up.

The database will start empty because this is using a production image, but the DB can be populated locally from R_test_data.sql using intelliJ or the postgres command line tools available in the running container e.g. with [INSERT POSTGRES COMMAND LINE TOOL FOR POPULATING THE DB LOCALLY HERE].
```

You can do this with `git commit --amend` in your terminal, although you may be using Intellij for this.

I've lost count of the number of times I've looked through the git history of other repositories to find the rationale for why a thing has changed, as it's often not obvious, so this would really help us in future.

Gweaton (approved):
Thanks for the update

PR #47, opened 2023-06-14
Temporarily disable E2E tests on CI

Gweaton:
## Changes in this PR

This disables end to end tests that run on CI as part of [our Playwright suite](https://github.com/ministryofjustice/hmpps-accredited-programmes-e2e).

These are going to be failing for a while as we're still building Find and sorting out dummy data, so removing them as a requirement for the build to pass feels like the quickest approach for now.

Once we've got a dev environment with more "stable" dummy data, we can re-enable these.


bell-pa (approved):
LGTM.

PR #48, opened 2023-06-14
(503) Add pact provider test for a single offering

Gweaton:
## Context

[503-add-new-pact-provider-test-for-a-single-offering
](https://trello.com/c/GTnZa5qw/503-add-new-pact-provider-test-for-a-single-offering)

## Changes in this PR

Adds a new Pact state to cover requests made for an individual offering with id `20f3abc8-dd92-43ae-b88e-5797a0ad3f4b`

### DIFF
--- a/src/test/resources/db/migration/R__test_data.sql
+++ b/src/test/resources/db/migration/R__test_data.sql
@@ -22,7 +22,7 @@ VALUES ('7fffcc6a-11f8-4713-be35-cf5ff1aee517', 'd3abc217-75ee-46e9-a010-368f302
        ('790a2dfe-7de5-4504-bb9c-83e6e53a6537', 'd3abc217-75ee-46e9-a010-368f30282367', 'BWN', 'nobody-bwn@digital.justice.gov.uk'),
        ('39b77a2f-7398-4d5f-b744-cdcefca12671', 'd3abc217-75ee-46e9-a010-368f30282367', 'BXI', 'nobody-bxi@digital.justice.gov.uk'),
 
-       ('ee20f564-4853-40b2-bae4-65dd7e0207fa', '28e47d30-30bf-4dab-a8eb-9fda3f6400e8', 'MDI', 'nobody-mdi@digital.justice.gov.uk'),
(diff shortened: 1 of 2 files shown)


src/test/resources/db/migration/R__test_data.sql
```diff
@@ -22,7 +22,7 @@ VALUES ('7fffcc6a-11f8-4713-be35-cf5ff1aee517', 'd3abc217-75ee-46e9-a010-368f302
        ('790a2dfe-7de5-4504-bb9c-83e6e53a6537', 'd3abc217-75ee-46e9-a010-368f30282367', 'BWN', 'nobody-bwn@digital.justice.gov.uk'),
        ('39b77a2f-7398-4d5f-b744-cdcefca12671', 'd3abc217-75ee-46e9-a010-368f30282367', 'BXI', 'nobody-bxi@digital.justice.gov.uk'),
 
-       ('ee20f564-4853-40b2-bae4-65dd7e0207fa', '28e47d30-30bf-4dab-a8eb-9fda3f6400e8', 'MDI', 'nobody-mdi@digital.justice.gov.uk'),
+       ('20f3abc8-dd92-43ae-b88e-5797a0ad3f4b', '28e47d30-30bf-4dab-a8eb-9fda3f6400e8', 'OCC', 'nobody-iry@digital.justice.gov.uk'),
```
Gweaton, line 25: @bell-pa is the aim to keep things consistent in the `InMemoryCourseRepository` here, or is it ok if this is out of sync with it?

bell-pa: There's no need to keep them in sync now - InMemoryCourseRepository is only used in tests.  I anticipate moving the code somewhere else and removing this class at some point.

bell-pa (approved):
Have the values that you changed here stopped changing in the Pact contract now? 

LGTM

Gweaton:
> Have the values that you changed here stopped changing in the Pact contract now?
> 
> LGTM

Yep, the value for the Course and CourseOffering's ids are fixed now (https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/blob/b67c58fae7f8444b6f6d0db425bb0c7a412d6c36/server/data/courseClient.test.ts#L25).

We were mistakenly generating them with randomised ids before, but now they'll stay the same.

We'll only need to update these contracts if the shape of things change in future, which shouldn't be an issue until we start building "Refer".

PR #50, opened 2023-06-15
Update dependencies following recommendations from renovate bot

bell-pa:
Update dependency versions in `build.gradle.kt` and `Chart.yml` to values specified by latest renovate bot branches.


Gweaton:
Looks like we've got a merge commit here - can we remove it and update with rebase as instead please?

PR #55, opened 2023-06-20
489 course upload

bell-pa:
## Context

Trello 489 [Add endpoint for uploading data to the API](https://trello.com/b/YoxohIMj/accredited-programmes-rm)

Candidate data for the set of Courses (Accredited Programmes) and Course Offerings is currently held and revised in a Google Sheets document.  This document is composed of several spreadsheets, of which the ones titled Course, Course Prerequisite and Offering are to be used to populate the `course`, `offering`, `prerequisite` and `audience` PosgreSQL database tables used by the the hmpps-accredited-programmes-api application.

The ticket behind this PR (and others to follow) asks for  a simple, repeatable method of moving the spreadsheet data to the PostgreSQL database.  There are, perhaps, two steps to this process
1. Extract data from the spreadsheets in a form that is amenable to upload to the database
2. Transform that data and insert into the tables, with integrity checks if needed.

Google Sheets has an export feature that translates a spreadsheet in to one of a number of different file formats.  The most appropriate formats for this task being the Microsoft Excel spreadsheet format, the Open Document Foundation spreadsheet format and text files in the 'Comma Separated Values' (CSV) and 'Tab Separated Values' (TSV) formats, the simplest of these being the CVS and TSV formats.

The extract step from Google sheets could be automated using the API that Google provide for this purpose. However, given that the extract / transform / load process is unlikely to occur more than a few times a year and may at a later date be replaced by a more 'user friendly' method, the extra time and work involved in implementing this step is hard to justy at this stage.  Therefore, the spreadsheet data will be extracted as a manual step to three files containing comma separated data.  These files should then be uploaded to the hmpps-accredited-programmes-api using PUT requests to three new end-points:

`PUT /courses` accepts csv format data from the Courses spreadsheet
`PUT /courses/offerings` accepts csv format data from the Offering spreadheet
`PUT /courses/prerequisites` accepts csv format data from the Course Prerequisites spreadsheet.

Each PUT operations is idempotent. That is, repeated applications of the PUT request using the same (valid) csv file will always put the application into the same state.  (Ignoring the synthetic ids which are surrogates for the business keys, an artifact of the implementation, and do not exist in the spreadsheets).

[...]

PR #57, opened 2023-06-21
(489) Add endpoint for uploading Courses CSV

Gweaton:
## Context

This PR is a restructure of #55, with all the hard work done by @bell-pa

https://trello.com/c/E7dsAUpL/489-add-endpoint-for-uploading-data-to-the-api

Candidate data for the set of Courses (Accredited Programmes) and Course Offerings is currently held and revised in a Google Sheets document. This document is composed of several spreadsheets, of which the ones titled Course, Course Prerequisite and Offering are to be used to populate the course, offering, prerequisite and audience PosgreSQL database tables used by the the hmpps-accredited-programmes-api application.

## Changes in this PR

This adds a simple, repeatable method of moving the Courses spreadsheet data to the PostgreSQL database. 

Extract data from the spreadsheets in a form that is amenable to upload to the database

The extract step from Google sheets could be automated using the API that Google provide for this purpose. However, given that the extract / transform / load process is unlikely to occur more than a few times a year and may at a later date be replaced by a more 'user friendly' method, the extra time and work involved in implementing this step is hard to justify at this stage. Therefore, the spreadsheet data will be extracted as a manual step to three files containing comma separated data. These files should then be uploaded to the hmpps-accredited-programmes-api using PUT requests to three new end-points:

`PUT /courses` accepts csv format data from the Courses spreadsheet (introduced in this PR)
`PUT /courses/offerings` accepts csv format data from the Offering spreadsheet (to come in a future PR)
`PUT /courses/prerequisites` accepts csv format data from the Course Prerequisites spreadsheet (to come in a future PR)

Each PUT operation is idempotent. That is, repeated applications of the PUT request using the same (valid) csv file will always put the application into the same state (ignoring the synthetic ids which are surrogates for the business keys, an artifact of the implementation, and do not exist in the spreadsheets).

The textual 'nesting' of the /offerings and /prerequisites end-points within the /courses end point is meant to indicate that the data represented by these end-points is dependent upon the courses end-point data. That is, the life-cycle of the offerings and pre-requisites for a course are dependent upon and wholly contained by the life-cycle of the course. Removing a course also removes its offerings and prerequisites. Therefore, a 'PUT /courses' replaces all course data and removes all offerings and prerequisites data.

This PR is the first of three. It represents the set of changes needed for an 'MVP' implementation of the `PUT /courses'. 

See commit messages for further details.


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesControllerTest.kt
```diff
@@ -17,12 +23,16 @@ import java.util.UUID
 @Import(JwtAuthHelper::class)
 class CoursesControllerTest(
   @Autowired val webTestClient: WebTestClient,
-  @Autowired val coursesService: CourseService,
+  @Autowired val inMemoryCourseRepository: InMemoryCourseRepository,
```
Gweaton, line 26: I think this change (and all changes to this file other than the new test for `coursesPut`) should come in its own commit, or at the very least we should document why it's changed in this commit if it's necessary for the Put endpoint - I don't know if it's directly related. @bell-pa what's the reason for this change?

bell-pa: CourseControllerTest used to use an injected CourseSerivce which in turn used InMemoryCourseRepository as a source of data. The controller now calls a service method that takes a parameter, and the method implementation is not trivial. I felt that it was better at this stage to switch to using a mock implementation of CourseService instead with the data for the mock responses being programmed from InMemoryRepository as before.

bell-pa (approved):
LGTM

PR #58, opened 2023-06-22
Update seeded prisons

Gweaton:
## Context

BWN wasn't and ID of a prison found in the Prison Register locally, so was always rendering empty offerings

## Changes in this PR

This PR replaces references to BWN in the local seed data with the IDs of prisons that were found in the local environment.


bell-pa (approved):
LGTM

PR #60, opened 2023-06-23
(538 & 539) Add bulk upload for Offerings and Prerequisites

bell-pa:
## Context

Trello: [Bulk upload of Prerequisites](https://trello.com/c/OTT2z5KI/538-endpoint-for-uploading-prerequisites)
Trello [Bulk upload Offerings](https://trello.com/c/S6orwLLN/539-endpoint-for-uploading-offerings)

The commits in this Pull Request provide bulk up-load functionality of accredited programmes data originating from Google spreadsheets produced by the team and exported to CSV files.

## Changes in this PR
This pull request first adds a PUT /courses/prerequisites end-point.  The end-point accepts data in CSV format.
Each row of data represents one course prerequisite. Each row should contain:
* name: The name of this prerequisite. Not unique.
* course: The name of the course to which this prerequisite applies.
* description: The description of the prerequisite.
* comments: Any additional information

Neither name, description or course are independently unique. However the business key <name, course, description> should be unique.

On upload the application deletes any stored prerequisites before adding the new values. Courses themselves are untouched. Each prerequisite must identify the course to which it applies. That is, the `course` value in the prerequisite row must match the `name` of a course.  The prerequisite is then added to the course.  When a match can't be found the prerequisite is ignored.  

Initially there is no feed back on the outcome of the upload.  Responses that may help with debugging are added in a commit at the end of this sequence.

Please refer to the commit messages for information on the purpose and content of each commit.

Once the prerequisite bulk upload is in place the process is repeated for course offerings: Add PUT /courses/offerings end-point that accepts a CSV format upload and replaces all offerings with those represented by this upload.
Each row should have
* **course** - the name of a course
* **prisonId** - A three letter code that identifies a prison (BWI, MDI, ...)
* **organisation** - ignored
* **contact email** - The email address of a contact at prisonId for this offering

Code-wise this is identical to the code already committed for the PUT /courses/prerequisites end-point.
Only the names of things (fields classes, tests) differ. So there's nothing new to explain here - comments on commits should be a sufficient guide.

Finally some diagnostic information is added to the response body of the prerequisite and offerings bulk upload end-points.


src/main/resources/static/api.yml
```diff
@@ -81,6 +82,31 @@ paths:
               schema:
                   $ref: '#/components/schemas/Course'
 
+  /courses/offerings:
+    put:
+      tags:
+        - Course offerings bulk upload / replace
+      summary: Upload a CVS format file containing a full set of offerings data for a set of courses.
```
Gweaton, line 95: Typo in CSV (CVS) here

src/main/resources/static/api.yml
```diff
@@ -243,6 +269,22 @@ components:
...
+          type: string
+        organisation:
+          type: string
+        "contact email":
+          type: string
+        prisonId:
+          type: string
+      required:
+        - course
+        - prisonId
```
Gweaton, line 297: I think contact email should be required here - if we can't display a contact email at this stage of "Find", people can't refer people in prison onto the course. Are there examples where this is missing?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesControllerTest.kt
```diff
@@ -40,199 +41,229 @@ class CoursesControllerTest(
...
-    mockMvc.get("/courses") {
-      accept = MediaType.APPLICATION_JSON
-      header(AUTHORIZATION, jwtAuthHelper.bearerToken())
-    }.andExpect {
-      status { isOk() }
-      content {
-        this.contentType(MediaType.APPLICATION_JSON)
-        this.json(
-          """
+  @Nested
```
Gweaton, line 44: Would you mind adding a line to the commit message that this commit also restructures the `coursesControllerTest` to group tests by function?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CsvTestData.kt
```diff
@@ -267,6 +268,247 @@ object CsvTestData {
...
+      OfferingRecord(course = "Identity Matters", prisonId = "GTI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "ISI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "LLI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "MRI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "SLI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "SNI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "MTI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "WRI"),
+      OfferingRecord(course = "Identity Matters", prisonId = "WMI"),
+      OfferingRecord(course = "Kaizen (sexual offences)", prisonId = "ASI"),
```
Gweaton, line 351: I wonder if we need to do this by some sort of `id` here - the Course name will just be `Kaizen` but the `(sexual offences)` will be the audience so we'll need a way of distinguishing between two courses with the same `name` but different `audience`s... Alternatively we could use `course` and `audience(s)` as a composite foreign key here. 

I've realised this also applies to prerequisites so we'll need to apply the same approach there. Any thoughts on what might be easier? Happy to help update the CSV for this.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/CourseService.kt
```diff
@@ -54,7 +54,18 @@ class CourseService(
...
 
-  fun replaceAllOfferings(offeringRecords: List<OfferingRecord>) {
-    offeringRecords.size
+  fun replaceAllOfferings(replacements: List<OfferingRecord>) {
+    val allCourses = courseRepository.allCourses()
+    clearOfferings(allCourses)
+    val coursesByName = allCourses.associateBy(CourseEntity::name)
+    replacements.forEach { record ->
+      coursesByName[record.course]?.run {
+        offerings.add(Offering(organisationId = record.prisonId, contactEmail = record.contactEmail ?: ""))
```
Gweaton, line 71: As mentioned earlier, `contactEmail` should always be present. I know it's not in the current CSV, so how about we log it in the same way we doing for prerequisites for which a matching course can't be found?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesIntegrationTest.kt
```diff
@@ -189,6 +186,8 @@ class CoursesIntegrationTest
       .bodyValue(CsvTestData.prerequisitesCsvText)
       .exchange()
       .expectStatus().is2xxSuccessful
+      .expectBody()
```
Gweaton, line 189: Ideally this would come in its own commit, or one that refactors things - unless I'm misunderstanding, is there a reason it's been added with the "Provide feedback on unmatched offerings and prerequisites" commit?

src/main/resources/static/api.yml
```diff
@@ -48,12 +48,18 @@ paths:
           'text/csv;charset=UTF-8':
             schema:
               type: array
-              title: 'coursePrerequisites'
               items:
                 $ref: '#/components/schemas/PrerequisiteRecord'
       responses:
-        204:
+        200:
```
Gweaton, line 54: Let's add some more context to the commit message as to why this is now a 200 response code.

Gweaton (commented):
Thanks for this - the approach looks good to me, have added a couple of minor comments.

One larger thing we need to figure out is how we should represent courses with the same name but different audiences. We can probably do that as another PR that builds on this work when we've come to a decision on this, which we may need to discuss with Louise.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesIntegrationTest.kt
bell-pa: Each commit builds on the previous one.  I could not have created a separate branch from main for this commit because it depends upon all the commits leading up to this one.
It might help if every commit were considered a release in its own right, each worthy of its own PR and merge to main. This, after all is what continuous integration means.  If you think that all the previous commits are OK then all of them are going to be merged to main. I don't think it makes any difference if you do them one at a time or all at once.  It would certainly be simpler to merge (or ideally rebase) the whole lot on to main in one go and the outcome is the same.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CsvTestData.kt
bell-pa: Seems to me that these are different courses that happen to have the same name. 
If HMPPS want to give the same name to different courses they're probably going to regret it.
We can't use audience(s) as part of a composite key because audience is separate entity in a many-to-many relationship with course. You could could either use different course names - Kaizen for sexual offences.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesIntegrationTest.kt
Gweaton: I think there might be some crossed wires here - I'm referring to "commit" rather than "branch" - this would be a separate commit, on the same branch.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/CourseService.kt
bell-pa: We could.  The presence or lack of an e-mail address is very obvious to anyone working with the spreadsheets and it is easy to remedy.  Do we really need extra checking here? If so then let's do this as a separate story/card.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CsvTestData.kt
Gweaton: They're the same courses, but different "strands" that we're representing with the `audience` fields (see designs [here](https://www.figma.com/proto/ZGMNCyn4uzmUxFQdmAiQQq/Accredited-Programmes-Prototype?type=design&node-id=1309-8923&scaling=min-zoom&page-id=892%3A7020&starting-point-node-id=1309%3A9318)) for how they're displayed.

Could you elaborate on the following? Could we not use the list of audience IDs and the course's name
 to identify it here?
> We can't use audience(s) as part of a composite key because audience is separate entity in a many-to-many relationship with course.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/CourseService.kt
Gweaton: I suppose, but if we're not worried about the presence of lack of an email address, do we need a default empty string value for it?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesIntegrationTest.kt
Gweaton: I'll accept updating the commit message here with a bit more detail as to why these tests have been updated, rather than splitting it out into a new commit to help get this over the line.

bell-pa: Ah.  Added .expectBody().jsonPath(... because prior to this commit there was no body.  Now there is, and because a number of the prerequisites fail to match a course by name the response body array has 141 entries, one for each mismatch. Line 199 asserts that 94 prerequisites were matched to courses as before.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/CoursesControllerTest.kt
bell-pa: Done

src/main/resources/static/api.yml
bell-pa: Will fix after merge to main.

src/main/resources/static/api.yml
bell-pa: Will adjust after merge to main.

Gweaton (approved):
Feel free to merge after disabling those Pact tests on CI

PR #62, opened 2023-06-27
(550) Create codeql.yml config

Gweaton:
## Context

https://trello.com/c/HFbiMKKr/550-add-codeql-for-vulnerability-scanning

CodeQL will scan our code for any vulnerabilities with GitHub Actions.

Unfortunately there's not yet automated configuration for Kotlin as there is for TypeScript (which we've used on the UI repo), so we have to add our own config file.

## Changes in this PR

Adds CodeQL config file to be run against `main` branch. This was mostly generated automatically when adding CodeQL in GitHub but I've made a few tweaks to have it line up with the same config used in the Approved Premises team [here](https://github.com/ministryofjustice/hmpps-approved-premises-api/blob/main/.github/workflows/codeql.yml)

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -31,9 +31,16 @@ jobs:
         languages: 'java'
         # Use only 'java' to analyze code written in Java, Kotlin or both
 
-    # Autobuild attempts to build any compiled languages (C/C++, C#, Go, Java, or Swift).
-    - name: Autobuild
-      uses: github/codeql-action/autobuild@v2
+    - name: Set up JDK 19
+      uses: actions/setup-java@v1
+      with:
+        java-version: 19
+
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+
+    - name: Build with Gradle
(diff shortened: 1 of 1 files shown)


bell-pa (approved):
LGTM
Noted that Kotlin code requires language 'java', not 'kotlin'.

.github/workflows/codeql.yml
```diff
@@ -31,9 +31,16 @@ jobs:
...
+    - name: Set up JDK 19
+      uses: actions/setup-java@v1
+      with:
+        java-version: 19
+
+    - name: Grant execute permission for gradlew
+      run: chmod +x gradlew
+
+    - name: Build with Gradle
+      run: ./gradlew build -x test
```
Gweaton, line 43: @bell-pa do you think this is an appropriate way of building the code for scanning? I've disabled the tests here as they're running on CI and we just need the code to be built to scan for vulnerabilities, but let me know if there's a better approach.

Running the build this way was also resulting in some failing tests, not sure why 🤷 

Anyway, this seems to work!

PR #69, opened 2023-07-03
533 Add identifier to course

bell-pa:
## Context

Trello [553 Modify bulk course upload to store course `identifier` in the database](https://trello.com/c/saKu5vkd/553-modify-bulk-course-upload-to-store-course-identifier-in-the-database)

## Changes in this PR
Introduced an `identifier` column to the `course` table and an identifier column to the `PUT /courses end-point `used for bulk upload of course data from a CSV file.  Modified the code so that the identifier values in the uploaded CSV are stored in the `course.identifier` column in the db.

## Release checklist

### Post-merge

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/NewCourse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/NewCourse.kt
@@ -1,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain
+
+data class NewCourse(
+  val name: String,
+  val description: String,
+  val audience: String,
+  val alternateName: String? = null,
+)
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/NewCourse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/NewCourse.kt
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain
 data class NewCourse(
   val name: String,
   val description: String,
+  val identifier: String,
   val audience: String,
   val alternateName: String? = null,
 )
--- a/src/main/resources/seed/db/migration/R__Seed_Data.sql
+++ b/src/main/resources/seed/db/migration/R__Seed_Data.sql
@@ -11,10 +11,10 @@ VALUES ('7fffcc6a-11f8-4713-be35-cf5ff1aee517', 'Sexual violence'),
        ('7fffcc6a-11f8-4713-be35-cf5ff1aee510', 'Violent offence'),
        ('7fffcc6a-11f8-4713-be35-cf5ff1aee511', 'Intimate partner violence');
 
-INSERT INTO course(course_id, name, description, alternate_name)
-VALUES ('d3abc217-75ee-46e9-a010-368f30282367', 'Lime Course', 'Explicabo exercitationem non asperiores corrupti accusamus quidem autem amet modi. Mollitia tenetur fugiat quo aperiam quasi error consectetur. Fugit neque rerum velit rem laboriosam. Atque nostrum quam aspernatur excepturi laborum harum officia eveniet porro.', 'LC'),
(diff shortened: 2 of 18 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/NewCourse.kt
```diff
@@ -3,6 +3,7 @@ package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain
 data class NewCourse(
   val name: String,
   val description: String,
+  val identifier: String,
```
Gweaton, line 6: I think there's a typo in the commit message here - you've used `identity` rather than `identifier` - I know, technically, the word has the same meaning in this case, but might be clearer to use `identifier` if you wouldn't mind rewording it.

src/main/resources/static/api.yml
```diff
@@ -223,6 +223,8 @@ components:
           type: string
         alternateName:
           type: string
+        identifier:
```
Gweaton, line 226: do we need to make this `required`?

src/main/resources/seed/db/migration/R__Seed_Data.sql
```diff
@@ -11,10 +11,10 @@ VALUES ('7fffcc6a-11f8-4713-be35-cf5ff1aee517', 'Sexual violence'),
...
 
-INSERT INTO course(course_id, name, description, alternate_name)
-VALUES ('d3abc217-75ee-46e9-a010-368f30282367', 'Lime Course', 'Explicabo exercitationem non asperiores corrupti accusamus quidem autem amet modi. Mollitia tenetur fugiat quo aperiam quasi error consectetur. Fugit neque rerum velit rem laboriosam. Atque nostrum quam aspernatur excepturi laborum harum officia eveniet porro.', 'LC'),
-       ('28e47d30-30bf-4dab-a8eb-9fda3f6400e8', 'Azure Course', 'Similique laborum incidunt sequi rem quidem incidunt incidunt dignissimos iusto. Explicabo nihil atque quod culpa animi quia aspernatur dolorem consequuntur.', 'AC++'),
-       ('1811faa6-d568-4fc4-83ce-41118b90242e', 'Violet Course', 'Tenetur a quisquam facilis amet illum voluptas error. Eaque eum sunt odit dolor voluptatibus eius sint impedit. Illo voluptatem similique quod voluptate laudantium. Ratione suscipit tempore amet autem quam dolorum. Necessitatibus tenetur recusandae aliquam recusandae temporibus voluptate velit similique fuga. Id tempora doloremque.', null);
+INSERT INTO course(course_id, name, identifier, description, alternate_name)
```
Gweaton, line 14: Do we need to add these seeded values to `src/test/resources/db/migration/R__test_data.sql` too, or is nothing failing yet?

Gweaton (commented):
Looks good to me, thanks for this - just a couple of small things/questions

src/main/resources/static/api.yml
bell-pa: I think a later commit makes this required.

src/main/resources/seed/db/migration/R__Seed_Data.sql
bell-pa: Yes,  see final commit.

src/main/resources/static/api.yml
Gweaton: oops, sorry, you're right!

src/main/resources/seed/db/migration/R__Seed_Data.sql
Gweaton: ah that makes sense. Although the final commit (https://github.com/ministryofjustice/hmpps-accredited-programmes-api/pull/69/commits/b54b5fc2f867fac1b7337bc4aa1f6d6b9e1aa60f) appears to only update the seed data, rather than the test data, unless I'm missing something.

Gweaton: but as you've said, if it's not necessary, that's ok

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/NewCourse.kt
bell-pa: You're right, I got mixed up in the code as well, then realised what I'd done.  I think the code is all correct (it must be otherwise it wouldn't compile and pass the tests).  Clearly I used the wrong term in the commit message.

Gweaton (approved):

PR #78, opened 2023-07-04
Eliminate inefficient Hibernate N+1 queries

bell-pa:
## Context

Invoking some api end-points results in Hibernate running 'N+1' queries to populate courses with their `audiences` and `prerequisites` collections.  Although the datasets are small this is still inefficient and is easily fixed.

## Changes in this PR

* Disable the Spring JPA 'open session in view' setting. 
* Set the fetch type for course.prerequisites, course.offerings and course.audiences to `FetchType.SUBSELECT `
* Selectively initialise collections in `JpaCourseRepository`

## Release checklist

### Post-merge

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -26,7 +26,7 @@ spring:
 
   jpa:
     database-platform: org.hibernate.dialect.PostgreSQLDialect
-    open-in-view: true # TODO: Hide the domain model from the view layer and stop using open-in-view.
+    open-in-view: false
     generate-ddl: false
     hibernate:
       ddl-auto: none
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,6 +46,10 @@ java {
   toolchain.languageVersion.set(JavaLanguageVersion.of(19))
 }
 
+kotlin {
+  kotlinDaemonJvmArgs = listOf("-Xmx486m")
+}
+
 tasks {
   withType<KotlinCompile> {
     kotlinOptions {
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/CourseEntity.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/CourseEntity.kt
@@ -11,6 +11,8 @@ import jakarta.persistence.JoinColumn
 import jakarta.persistence.JoinTable
 import jakarta.persistence.ManyToMany
 import jakarta.persistence.Table
+import org.hibernate.annotations.Fetch
+import org.hibernate.annotations.FetchMode.SUBSELECT
 import java.util.*
 
 @Entity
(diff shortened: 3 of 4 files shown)


src/main/resources/application.yml
```diff
@@ -26,7 +26,7 @@ spring:
 
   jpa:
     database-platform: org.hibernate.dialect.PostgreSQLDialect
-    open-in-view: true # TODO: Hide the domain model from the view layer and stop using open-in-view.
+    open-in-view: false
```
Gweaton, line 29: Just for my own understanding, why have we disabled this? From a quick google, it looks like it's a bit of an anti-pattern to have on, but none of the articles are very Kotlin-beginner friendly...

Gweaton (approved):

src/main/resources/application.yml
bell-pa: When Spring handles a request that involves interaction with Hibernate it creates and manages a Hibernate (JPA) Session which lasts for the duration of the request. Spring should also initiate and manage a database transaction following instructions in `@Transaction` annotations in the code.  

In this application there is a `@Transaction` annotation on the CourseService class.  Whenever the thread of control enters one of the methods of this class a database transaction is started for that thread. When the method is exited the transaction is committed, or if certain exceptions are thrown or the Tx is marked for rollback then the Tx is rolled back.

The Hibernate Session is the entry point to the Hibernate api. The session is associated with the request (actually the thread handling the request). It contains a cache, and manages lazy loading, query plans, queues database operations and so on.

The CourseEntity composite has three collections: `offerings`, `prerequisites` and `audiences`.  When the Hibernate session is asked to retrieve an entity from the database (`findAll` or `findById` for example) it may not necessarily do so immediately.  When it does fetch data from the db then various kinds of configuration: annotations, fetch graphs, the way the query is written, decides which parts of the composite are populated.

How does this work?  Hibernate replaces parts of the object graph with proxies. Visiting a proxy, that is attempting to fetch data from or put data to the proxy triggers hibernate to fetch the data behind it using various strategies and this data is then supplied to the client.

The CourseEntity collections are all configured for lazy fetch. Prior to the changes in this PR, invoking the `GET /courses` end-point worked like this:
* Call the `CourseService.allCourses` method. Transaction begins on method entry.  
* Service calls `courseRepository.allCourses()` and Hibernate returns a list of `CourseEntity` in which the `CourseEntity` objects are populated but the collections are represented by proxies and remain unpopulated.
* `CourseService.allCourses` returns a list of CourseEntity to the controller.  Transaction is committed.
* The controller returns the partially populated list of CourseEntity.
* Spring aspects intercept the return, walk the list and convert it to JSON.  In doing so the audiences and prerequisites collections on each entity are visited in turn.  Each visit causes the corresponding proxy to trigger a db query to fetch the collection of audience or prerequisite represented by the proxy.  Each query is run in a separate auto-commit transaction.  So Hibernate executes (2n + 1) db queries, all but one outside the transaction boundary to produce this result.

The open session in view configuration option is the setting that results in the collections being populated in the controller code.  When this setting is disabled the session is closed when the transaction is committed. Any unpopulated collections or object remain unpopulated and accessing them throws an exception.

The other two parts of the optimisation are
* Specifying a fetch strategy for the collections.  I've set this to `@Fetch(SUBSELECT)` for all collections. The collections are still fetched lazily, but visiting one collection in the result set (prerequisites on one CourseEntity for example) results in Hibernate running one query to fetch and populate all the prerequisites collections in the result set.
* Asking Hibernate to populate the collections before the `CoursesJpaRepository` returns the list of CourseEntity or single CourseEntity.  There are other ways to do this, but the way I chose is simple and quick.
Now the parts of the list of CourseEntity visited by the controller's JSON renderer are fully populated and the end-points work correctly without relying on open-session-in-view.

PR #91, opened 2023-08-01
614 Course, prerequisite and offering updates

bell-pa:
## Context
Trello [Investigation/spike: how should we update data in the service?](https://trello.com/c/sooyolH2/614-investigation-spike-how-should-we-update-data-in-the-service)

## Changes in this PR

To allow rapid development of a mechanism for uploading data from spreadsheets directly into the api this mechanism erased all data and replaced it with new each time a valid CSV format file was uploaded to the `PUT /courses` end-point.

This worked well for the initial release but will not be good-enough for the longer term.  Bulk delete and replacement will not work when entities are referred to by other long-lived entities such as Referrals because each bulk updated changes every surrogate key even when entities identified by the keys do not change. Logically this doesn't make sense and practically it doesn't work because it breaks the primary and foreign key constraints on the relationship between, for example, Referral and Offering. (When Referrals are added to the API).

This PR re-works `CourseService` so that a bulk upload updates existing entities in place, adds new entities and withdraws (soft deletes) entities instead of deleting them.  Now the attributes of entities are updated while the entity itself remains in-situ and retains its surrogate key (id).  Even if an entity (Course, Offering) is inadvertently deleted it can be restored in a subsequent update by reverting the entity's withdrawn flag.

### Post-merge


bell-pa:
Withdrawn - out of date and replaced by a newer sequence of PRs.

PR #116, opened 2023-09-18
614 update offerings instead of replacing

bell-pa:
## Context
Currently the only way to update Courses, Offerings, Prerequistes etc is using the bulk update mechanism.
This consists of POSTing CSV files to the `PUT /courses`, `PUT /offerings` and `PUT /prerequisites` endpoints.

Before this PR Offerings were updated by:
* Deleting all offerings
* Inserting an Offering for each valid record in the uploaded offerings CSV file.

This becomes a problem when referrals are linked to offerings because replacing an Offering changes the Offering's ID.
The solution is to add a soft delete option to Offerings (done) and then update Offerings in place.
Matches are determined by Course identifier and OrganistionId (PrisonId).

This is what happens when a set of offering records is uploaded:
* The new offerings are grouped by CourseId.
* The current offerings and the new offerings for each course are compared
* Every offering that matches an update is updated in place and its withdrawn flag set false
* Every new offering that has no match is added the course's offerings
* Every  course offering that is absent from the new offerings is withdrawn

Trello: [Update Courses, Prerequisites, Offerings instead of replacing](https://trello.com/c/sooyolH2/614-add-scripts-for-updating-course-offering-data-in-the-service)

## Changes in this PR

* The `CourseService` `replaceAllOfferings` method has been renamed to `updateOfferings`. 
* The data class that represents updates has been renamed from `NewOffering` to `OfferingUpdate`
* The logic in the `CourseService` `updateOfferings` method to perform updates in place as outlined above
* `CourseService` tests updated and extended to assert that the new behaviour is correct

### Post-merge

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
@@ -75,28 +75,51 @@ class CourseService(
     courses.forEach { it.prerequisites.clear() }
   }
 
-  fun replaceAllOfferings(replacements: List<NewOffering>): List<LineMessage> {
+  fun updateOfferings(updates: List<OfferingUpdate>): List<LineMessage> {
     val allCourses = courseRepository.allCourses()
-    clearOfferings(allCourses)
     val coursesByIdentifier = allCourses.associateBy(CourseEntity::identifier)
-    replacements.forEach { record ->
-      coursesByIdentifier[record.identifier]?.run {
-        addOffering(
-          Offering(
-            organisationId = record.prisonId,
-            contactEmail = record.contactEmail
-              ?: "",
-            secondaryContactEmail = record.secondaryContactEmail,
-          ),
-        )
+    val desiredOfferingsByCourseIdentifier = updates.groupBy(OfferingUpdate::identifier)
+    coursesByIdentifier.forEach { (courseIdentifier, course) ->
+      updateOfferingsForCourse(course, desiredOfferingsByCourseIdentifier[courseIdentifier] ?: emptyList())
+    }
+
+    return contactEmailWarnings(updates) + unmatchedCourseErrors(updates, coursesByIdentifier)
+  }
+
+  private fun updateOfferingsForCourse(course: CourseEntity, desiredOfferings: List<OfferingUpdate>) {
+    val offeringsByOrganisationId = course.offerings.associateBy(Offering::organisationId)
+    val updatesByPrisonId = desiredOfferings.associateBy(OfferingUpdate::prisonId)
+
+    val toAdd = updatesByPrisonId.keys - offeringsByOrganisationId.keys
+    toAdd.forEach {
+      val update = updatesByPrisonId[it]
+      course.addOffering(
+        Offering(
+          organisationId = it,
+          contactEmail = update?.contactEmail ?: "",
+          secondaryContactEmail = update?.secondaryContactEmail,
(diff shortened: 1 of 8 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
```diff
@@ -75,28 +75,51 @@ class CourseService(
...
+    }
+
+    return contactEmailWarnings(updates) + unmatchedCourseErrors(updates, coursesByIdentifier)
+  }
+
+  private fun updateOfferingsForCourse(course: CourseEntity, desiredOfferings: List<OfferingUpdate>) {
+    val offeringsByOrganisationId = course.offerings.associateBy(Offering::organisationId)
+    val updatesByPrisonId = desiredOfferings.associateBy(OfferingUpdate::prisonId)
+
+    val toAdd = updatesByPrisonId.keys - offeringsByOrganisationId.keys
```
Gweaton, line 93: Is there a more explicit variable name we could use here? Are these Offering IDs or Prison IDs? Or both?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
```diff
@@ -75,28 +75,51 @@ class CourseService(
...
+    toAdd.forEach {
+      val update = updatesByPrisonId[it]
```
Gweaton, line 95: should this be `addition`?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
```diff
@@ -75,28 +75,51 @@ class CourseService(
...
-        )
+    val desiredOfferingsByCourseIdentifier = updates.groupBy(OfferingUpdate::identifier)
+    coursesByIdentifier.forEach { (courseIdentifier, course) ->
+      updateOfferingsForCourse(course, desiredOfferingsByCourseIdentifier[courseIdentifier] ?: emptyList())
+    }
+
+    return contactEmailWarnings(updates) + unmatchedCourseErrors(updates, coursesByIdentifier)
+  }
+
+  private fun updateOfferingsForCourse(course: CourseEntity, desiredOfferings: List<OfferingUpdate>) {
```
Gweaton, line 89: This method is doing a few things, I wonder if it would be more easily scannable for future readers if we split it into three smaller private methods:
`addNewOfferingsToCourse`
`updateCourseOfferings`
`withdrawOfferingsFromCourse`

What do you think?

Gweaton (commented):
Looks good - just have a couple of comments that would help readability.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
bell-pa: We upload prison Ids, but internally they're called organisation ids.

Gweaton: Oh yeah. Could we call this `organisationIdsToAdd`?

Gweaton (approved):
Approving this, but let's do a bit of refactoring once it's merged.

PR #118, opened 2023-09-19
614 update courses instead of replacing

bell-pa:
## Context
Currently the only way to update Courses, Prerequistes etc is using the bulk update mechanism.
This consists of POSTing CSV files to the `PUT /courses`  and `PUT /prerequisites` endpoints.

Before this PR courses were updated by:
* Deleting all courses (and prerequisites, offerings, audiences)
* Inserting a course and associated audiences for each valid record in the uploaded courses CSV file.

This becomes a problem when referrals are linked to offerings.  Referrals have a foreign key to an offering id which will be violated when the offerings are deleted.
The solution is to add a soft delete option to Offerings and Courses (done) and then update Offerings (done) and Courses in place.
Courses matches are determined by Course identifier.

This is what happens when a set of course records is uploaded:
* The set of audience values in the upload is used to update the audience table: Any missing values are added to the table.
* The new courses are grouped by identifier (not id).
* The existing courses are grouped by identifer (not id).
* The sets of new and old identifiers are compared
* Every existing course that matches an update is updated in place and its withdrawn flag set false
* Every new course that has no match is added
* Every  course offering that is absent from the new courses is withdrawn

Trello: [Update Courses, Prerequisites, Offerings instead of replacing](https://trello.com/c/sooyolH2/614-add-scripts-for-updating-course-offering-data-in-the-service)

## Changes in this PR

* The `CourseService` `replaceAllCourses` method has been renamed to `updateCourses`. 
* The data class that represents updates has been renamed from `NewCourse` to `CourseUpdate`
* The logic in the `CourseService` `updateCourses` method has been changed to perform updates in place as outlined above
* `CourseService` tests updated and extended to assert that the new behaviour is correct
* `CoursesIntegrationTest` for `PUT /courses` has been updated to demonstrate that courses can be soft deleted, restored and updated without changing assigned course ids.
* Added a NOT NULL constraint to table column course.identifier
* Added a UNIQUE constraint to table column course.identifier
* Added identifiers to courses in the `R__test_data.sql` dataset

### Post-merge


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
```diff
@@ -25,22 +25,57 @@ class CourseService(
...
-        identifier = it.identifier,
-        description = it.description,
-        alternateName = it.alternateName,
-        audiences = audienceStrings(it.audience).mapNotNull { audienceName -> allAudiences[audienceName] }.toMutableSet(),
-        referable = it.referable,
-      )
+    val coursesByIdentifier = courseRepository.allCourses().associateBy(CourseEntity::identifier)
+    val courseDataByIdentifier = courseData.associateBy(CourseUpdate::identifier)
+
+    val toAdd = courseDataByIdentifier.keys - coursesByIdentifier.keys
```
Gweaton, line 34: as mentioned in my review on your other PR, it'd be good to be a bit more explicit about what these variables are holding in the variable name

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
```diff
@@ -25,22 +25,57 @@ class CourseService(
...
+    val toUpdate = coursesByIdentifier.keys.intersect(courseDataByIdentifier.keys)
+    val toWithdraw = coursesByIdentifier.keys - courseDataByIdentifier.keys
+
+    toAdd.mapNotNull {
```
Gweaton, line 38: again, like my previous PR, I wonder if it's worth us extracting this behaviour into separate `add`, `update` and `withdraw` private functions, a bit like we have for `updateAudiences` as having these smaller functions would help improve readability.

Gweaton (commented):
This looks good to me, just a couple of small comments (similar vein to my previous PR review!)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
bell-pa: I agree.  I'd like to do this in a follow-up refactoring PR. Its simpler.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/domain/CourseService.kt
bell-pa: I agree.  I'd like to do this in a follow-up refactoring PR. Its simpler.

Gweaton (approved):
Approving this, but let's do a bit of refactoring once it's merged.

PR #120, opened 2023-09-20
614 add bulk csv download endpoints for round trips

bell-pa:
## Context

Trello: 
* [Add scripts for updating course data](https://trello.com/c/sooyolH2/614-add-scripts-for-updating-course-offering-data-in-the-service)
* [Mark non-referrable courses as non-referrable](https://trello.com/c/DygFUaSW/837-mark-non-referable-courses-as-non-referable-api)

The API contains PUT endpoints for bulk add/update/delete operations on Courses, Offerings, Audience and Prerequisites.
This PR adds corresponding GET endpoints for downloading a CSV format representation of Courses, Offerings, Audience and Prerequisites.  The documents from these endpoints are compatible with their corresponding PUT endpoints so that the data can be round-tripped.  

These bulk downloads may be saved as CSV format files and modified in spreadsheet editing programmes such as Microsoft Excel or Google Sheets.  The modified files can then be uploaded to the API thereby applying the changes to the course related data stored in the API.

## Changes in this PR
* Add an implementation of the `writeInternal` method of the  `CsvHttpMessageConverter` class. Doing so makes it possible for Controller endpoints to serve CSV format documents from `List<T>` where T is any flat DTO-like class. (probably a data class)
* Extend `api.yml` to add the GET endpoints for CSV downloads
* Implement the `GET /courses` endpoint
* Implement the `GET /offerings` endpoint and move the existing `PUT /courses/offerings` endpoint to `PUT /offerings`
* Implement the `GET /courses/prerequisites` endpoint
* Add scripts to the `/script/local-scripts` directory that invoke the CSV download endpoints
* Update the README.md in `/script/local-scripts`

### Post-merge

### DIFF
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -19,6 +19,8 @@ paths:
                 type: array
                 items:
                   $ref: '#/components/schemas/Course'
+
+  /courses/csv:
     put:
       tags:
         - Courses
@@ -36,8 +38,6 @@ paths:
           description: Successful update
         400:
           description: Bad input
-
-  /courses/csv:
     get:
       tags:
         - Courses
@@ -52,7 +52,7 @@ paths:
                 items:
                   $ref: '#/components/schemas/CourseRecord'
 
-  /courses/prerequisites:
+  /courses/prerequisites/csv:
     get:
       tags:
         - Course Prerequisites
@@ -142,7 +142,7 @@ paths:
                 items:
                   $ref: '#/components/schemas/CourseOffering'
 
-  /offerings:
+  /offerings/csv:
     get:
       tags:
         - Course Offerings
--- a/src/main/resources/static/api.yml
+++ b/src/main/resources/static/api.yml
@@ -119,11 +119,34 @@ paths:
               schema:
                   $ref: '#/components/schemas/Course'
 
-  /courses/offerings:
+  /courses/{courseId}/offerings:
     get:
       tags:
         - Course Offerings
-      summary: Download a CSV format representation of those offerings that are not withdrawn.
+      summary: List all offerings for a course
+      parameters:
+        - name: courseId
+          in: path
+          description: A course identifier
+          required: true
+          schema:
+            type: string
+            format: uuid
+      responses:
+        200:
+          description: successful operation
+          content:
+            'application/json':
+              schema:
+                type: array
+                items:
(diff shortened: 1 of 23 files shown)


src/main/resources/static/api.yml
```diff
@@ -37,7 +37,35 @@ paths:
         400:
           description: Bad input
 
+  /courses/csv:
```
Gweaton, line 40: So if I'm understanding correctly, we can't have:
`get courses/` with `text/csv` returns the CSV
`get courses/` with `application/json` returns the json response

I wonder if we should therefore have all of these endpoints be `/csv`, both for consistency of endpoints _and_ if in future we were to add another `/offerings` endpoint that would fetch all offerings, this would cause a clash.

What do you think?

Gweaton (commented):
This looks good to me - I've left one thought re: consistency of endpoints which I think would be good to address.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/course/restapi/CsvTestData.kt
```diff
@@ -20,7 +20,7 @@ object CsvTestData {
 
   val newCourseUpdates: List<CourseUpdate> = listOf(
     newCourse(name = "Becoming new me Plus", identifier = "BNM-SO", audience = "sexual offence", alternateName = "BNM+++"),
-    newCourse(name = "Becoming new me Plus", identifier = "BNM-VO", audience = "general violence offence", alternateName = "BNM+-", referable = false),
+    newCourse(name = "Becoming new me Plus", identifier = "BNM-VO", audience = "general violence offence, mystery offence", alternateName = "BNM+-", referable = false),
```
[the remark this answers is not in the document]
bell-pa: yes

src/main/resources/static/api.yml
bell-pa: There isn't a nice solution to this.  I don't object to adding the /csv suffix, but I'd rather the code generator did the right thing! Or... that I could work out how to get the code generator to do the right thing.

bell-pa: Done.

Gweaton (approved):
Thanks for making those changes

PR #130, opened 2023-09-28
862 Course participation audit fields - second step - manual rebase

bell-pa:
## Context

Trello: [Add additional programme history fields to API](https://trello.com/c/kWwW6vM1/862-add-additional-programme-history-fields-to-api)

## Changes in this PR
The card asks for an `addedBy` (required) text field (name/username?)  and a `createdAt` field to be added to programme history records returned from the API.

This is achieved by 
* Enabling JPA Auditing
* Adding an implementation of the `AuditorAware` interface. The implementation obtains a username from the Spring Security SecurityContext.  The Spring Security implementation extracts this username from the JWT token provided with the request.
* Adding `@EntityListeners(AuditingEntityListener::class` to the `CourseParticipation` to make it auditing aware
* Adding audit properties annotated with `@CreatedBy`,` @CreatedDate`, `@LastModifiedBy` and `@LastModifiedDate`. These are managed by the Spring Data auditing feature.

Repository and integration tests show that the username embedded in the request's JWT is persisted and later returned as the value of the `addedBy` property of responses containing JSON representations of `CourseParticipation` objects.

## Post-merge

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/common/jpa/ExtraJpaConfiguration.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/common/jpa/ExtraJpaConfiguration.kt
@@ -1,0 +1,25 @@
+package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.common.jpa
+
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.data.domain.AuditorAware
+import org.springframework.data.jpa.repository.config.EnableJpaAuditing
+import org.springframework.security.core.context.SecurityContextHolder
+import org.springframework.security.core.userdetails.UserDetails
+import java.util.Optional
+
+@Configuration
+@EnableJpaAuditing(modifyOnCreate = false)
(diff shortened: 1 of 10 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/common/jpa/ExtraJpaConfiguration.kt
```diff
@@ -0,0 +1,25 @@
...
+
+@Configuration
+@EnableJpaAuditing(modifyOnCreate = false)
+class ExtraJpaConfiguration {
+  @Bean
+  fun auditorAware() = AuditorAware {
+    Optional.ofNullable(
+      when (val principal = SecurityContextHolder.getContext().authentication?.principal) {
+        is String -> principal
+        is UserDetails -> principal.username
```
Gweaton, line 19: Out of interest, do we get the user's first and last name on the `principal` here? I'm now thinking the full name would be more user-friendly than the username (which often looks like `EG012345` in prod). If not, I'm assuming we'd need to make a request to the `manage-users-api` to fetch that?

Gweaton (approved):
I think this makes sense, I just had one question on usernames vs real names

PR #143, opened 2023-10-10
Use `Hibernate.getClass(this)` in entity `equals` methods 

bell-pa:
Hibernate `equals` methods should be written in a particular way so that they honour the contract for `equals`:
* If the compared objects are the same by reference then they are equal
* If the compared object is null then they are not equal.
* If both objects are instances of the same class according to the `Hibernate.getClass()` method and both ids are non-null and equal then the objects are equal, otherwise they are not.

Updates `CourseEntity`, `Offering` and `Referral`.

Also changed `CourseService` to use the inline version of the Set method `intersect`.

### Post-merge


nsmj63-dxw:
Closed for now as we look to further reduce our reliance on Hibernate functions, in favour of Kotlin `data class`es.

PR #150, opened 2023-10-19
#958 Add new `courses/course-names` endpoint

nsmj63-dxw:
## Context

> see [#958 on the Accredited Programmes Trello board.](https://trello.com/c/ZbZb6IUc/958-add-course-names-endpoint)

## Changes in this PR

The current way in which course names are retrieved do not account for duplicate named results. Whilst the front-end could feasibly deal with each individual request dynamically, by filtering the response from the standard `/courses/` endpoint, we felt it was more prudent to enshrine this sort of filtering functionality in the back-end, in preparation for (and therefore to remain consistent with) later planned tickets for filtering, sorting, and pagination.

In this case, rather than define a `/course-names` endpoint, which would have the OpenApi generator define an entirely new `CourseNamesController` flow, I have implemented this new endpoint under the existing `/courses/` URI string, thus `courses/course-names`.

In the future, if we ever needed to filter further, this endpoint can scope out to e.g. include a filtering term as a path parameter or request body. For now, however, I have kept things simple for the sake of easy integration.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/controller/CourseController.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/controller/CourseController.kt
@@ -75,4 +75,13 @@ constructor(
           .getAllOfferingsByCourseId(id)
           .map(OfferingEntity::toApi),
       )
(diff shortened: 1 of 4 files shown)


Gweaton (approved):
Looks good to me from a functional perspective

PR #152, opened 2023-10-20
#860 add new `additionalInformation` field to `Referral`

nsmj63-dxw:
## Context

> [see #860 on the Accredited Programmes Trello board.](https://trello.com/c/ktEadiIa/860-rename-reason-field-to-additionalinformation-in-the-data-model-and-api)

## Changes in this PR

We currently have a `reason` field in `Referral` which needs to be changed to `additionalInformation`.
This will be done in three steps:
- add the new `additionalInformation` field and deprecate the old `reason` field in the API (this PR).
- update the UI to make use of the new field and deprecate the old field.
- remove the legacy field from the API.

## Release checklist

### Pre-merge

### Post-merge


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/update/ReferralUpdate.kt
```diff
@@ -0,0 +1,8 @@
+package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.update
+
+data class ReferralUpdate(
```
Gweaton, line 3: Out of interest, is having these `Update`-style classes a common pattern? I know @anandelangovan1 had a couple of comments on this, and how we might want to use the `Referral` entity itself, but I may have misunderstood.

Gweaton (approved):
Looks good, just had one question

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/create/ReferralEntity.kt
```diff
@@ -26,7 +26,9 @@ data class ReferralEntity(
   val offeringId: UUID,
   val prisonNumber: String,
   val referrerId: String,
+  @Deprecated(message = "please use ReferralEntity.additionalInformation instead")
   var reason: String? = null,
+  var additionalInformation: String? = null,
```
Gweaton, line 31: Ah I've just realised this should be mandatory

Gweaton: But we should probably do this in the PR where we get rid of the old field.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/update/ReferralUpdate.kt
nsmj63-dxw: Mainly it's for flexibility. Creation might require certain fields to be mandatory, while updates might not. By having separate data classes for each of those functions, we can define different sets of required fields. Conversely, there might be fields that should not be updated once they're set. With separate classes, we're able to just omit those fields from the Update DTO, reducing our overheads. That said, I also like them for their relative simplicity, because they force us to be more explicit with the data we're passing to the API. It's much easier to see what a Create or an Update is actually doing when there are literally fewer fields you're allowed to pass in, and it stops us from needlessly passing in fields which don't need to be updated.

PR #161, opened 2023-11-14
1050 create prison api client

anandelangovan1:
## Context

https://trello.com/c/K3a1pdwz/1050-integrate-with-prison-api-m

## Changes in this PR

As part of this PR we are adding a prison-api client with 1 methods to fetch SentencesAndOffences by bookingId 
Setting up of the prison-api locally has been tricky and local testing is challenged by the lack of data. The data would need a bit more tweaking before before it can be used properly

This also includes the base class which makes the request to the external service making it extensible for future calls to other services. 

	 

## Release checklist

### Pre-merge

### Post-merge


nsmj63-dxw:
I think it'd be really useful for this PR to write a short explanation of the changes made, since this is one of the most important changes we're making to our service.

src/main/resources/application.yml
```diff
@@ -87,8 +87,17 @@ management:
       cache:
         time-to-live: 2000ms
 
+
+services:
+  prisons-api:
+    base-url: http://localhost:9094
```
nsmj63-dxw, line 93: Which of these is correct here? If one is needed for a different environment then should we move it in a different `application-x.yaml` file?

src/test/resources/application-test.yml
```diff
@@ -22,7 +22,7 @@ spring:
 
 hmpps:
   auth:
-    url: http://localhost:#WIREMOCK_PORT/auth
+    url: http://localhost:9090/auth
```
nsmj63-dxw, line 25: Double-check if we still need wiremock before making this change official.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/response/model/SentenceAndOffenceDetails.kt
```diff
@@ -1,22 +1,70 @@
 package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.response.model
 
+import jakarta.validation.constraints.NotNull
+import java.time.LocalDate
+
 class SentenceAndOffenceDetails {
```
nsmj63-dxw, line 6: These are huge classes. I was under the impression that we only needed a specific few fields from each external API. If we don't need all these fields then we should reduce the size of the projection DTO that we're defining.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/response/model/SentenceDetail.kt
```diff
@@ -1,55 +1,63 @@
 package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.response.model
 
+import jakarta.validation.constraints.NotNull
+import java.time.LocalDate
+
 class SentenceDetail {
```
nsmj63-dxw, line 6: Same as the last comment. This is a huge class whose details we are very unlikely to need more than a few fields from.

nsmj63-dxw (changes requested):
A few things I'd change here.
Don't forget to run `ktLintFormat` and make fixup commits, once you've made all the changes you need.

src/main/resources/application.yml
```diff
@@ -23,6 +23,16 @@ spring:
       resourceserver:
         jwt:
           jwk-set-uri: ${hmpps.auth.url}/.well-known/jwks.json
+      client:
+        registration:
+          prisons-api:
+            provider: hmpps-auth
+            client-id: approved-premises-api
```
Gweaton, line 30: This needs to match local HMPPS Auth credentials defined [here](https://github.com/ministryofjustice/hmpps-auth/blob/f2c661d8e7237e8c500682366ad87e0a0b0b6359/src/main/resources/db/dev/data/auth/V900_0__clients.sql#L34) and should be `hmpps-accredited-programmes-client-1` (`approved-premises` is another project, the one we've taken all of this from!)

src/main/resources/application.yml
```diff
@@ -77,4 +87,13 @@ management:
       cache:
         time-to-live: 2000ms
 
+
+services:
+  prisons-api:
```
Gweaton, line 92: Can we call this `prison-api` to match the API name - this is different from the approved premises project's naming, but I think they may have got it wrong...

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/client/PrisonsApiClient.kt
```diff
@@ -0,0 +1,22 @@
+package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.client
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.springframework.beans.factory.annotation.Qualifier
+import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.WebClient
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.response.model.SentenceAndOffenceDetails
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.response.model.SentenceDetail
+
+@Component
+class PrisonsApiClient(
```
Gweaton, line 10: Let's call this `PrisonApiClient` (remove the `s`) to match the name of [the service we're calling](https://github.com/ministryofjustice/prison-api). And for the `@Qualifier` below.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/config/WebClientConfiguration.kt
```diff
@@ -0,0 +1,67 @@
...
+    val service: OAuth2AuthorizedClientService = InMemoryOAuth2AuthorizedClientService(clients)
+    val manager = AuthorizedClientServiceOAuth2AuthorizedClientManager(clients, service)
+    val authorizedClientProvider = OAuth2AuthorizedClientProviderBuilder.builder()
+      .clientCredentials()
+      .build()
+    manager.setAuthorizedClientProvider(authorizedClientProvider)
+    return manager
+  }
+
+  @Bean(name = ["prisonsApiWebClient"])
```
Gweaton, line 38: If we rename the client, we'll want to remove the `s` on `prison` here.

helm_deploy/hmpps-accredited-programmes-api/values.yaml
```diff
@@ -33,6 +33,8 @@ generic-service:
     hmpps-accredited-programmes-api:
       APPINSIGHTS_INSTRUMENTATIONKEY: "APPINSIGHTS_INSTRUMENTATIONKEY"
       SENTRY_DSN: "SENTRY_DSN"
+      SPRING_SECURITY_OAUTH2_CLIENT_REGISTRATION_PRISONS-API_CLIENT-ID: "SYSTEM_CLIENT_ID"
```
Gweaton, line 36: I think the keys should be `SYSTEM_CLIENT_ID` and `SYSTEM_CLIENT_SECRET` to match what's in our kubernetes secrets, otherwise we'd need to add these in, right?

helm_deploy/values-dev.yaml
```diff
@@ -12,6 +12,7 @@ generic-service:
     APPLICATIONINSIGHTS_CONFIGURATION_FILE: applicationinsights.dev.json
     HMPPS_AUTH_URL: https://sign-in-dev.hmpps.service.justice.gov.uk/auth
     SENTRY_ENVIRONMENT: dev
+    SERVICES_PRISONS-API_BASE-URL: https://prison-api-dev.prison.service.justice.gov.uk
```
Gweaton, line 15: Can we just call this `PRISON_API_URL` to match the structure of `HMPPS_AUTH_URL` above? We'll need to do this across the three envs.

src/main/resources/application-local.yml
```diff
@@ -1,6 +1,6 @@
 oauth:
   client:
-    id: whereabouts-api-client
+    id: accredited-programmes-api
```
Gweaton, line 3: Does this work locally, or does it need to be `hmpps-accredited-programmes-client-1` as in `application.yml`?

Gweaton (commented):
This generally looks good, but there are a few naming issues and things that would be good to tidy up, particularly in terms of how we commit and document this for future readers. It was a big piece of work and a lot of investigation was required, so anything we can do to document our newfound knowledge and approach with PR descriptions and commit messages would help. Let's pair on this.

src/main/resources/application.yml
anandelangovan1: This is something which am still trying to figure out. We have `application.yaml`, `application-local.yaml`,`application-test.yaml` 

Trying to work out which yaml takes precedence

src/test/resources/application-test.yml
anandelangovan1: Good shout. 
Our docker compose points at 9090 for auth. Will check with George

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/response/model/SentenceAndOffenceDetails.kt
anandelangovan1: Again something to check I guess. I've taken the entire response from the prison-api. Couldn't find in the trello ticket on which fields were required and which weren't

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/config/SecurityConfiguration.kt
```diff
@@ -94,6 +102,56 @@ class AuthAwareTokenConverter : Converter<Jwt, AbstractAuthenticationToken> {
   }
 }
 
+@Configuration
+class AuthorizedClientServiceConfiguration(
+  @Value("\${log-client-credentials-jwt-info}") private val logClintCredentialsJwtInfo: Boolean,
```
Gweaton, line 107: `logCli[e]ntCredentialsJwtInfo`

src/main/resources/application-local.yml
Gweaton: Having tested this yesterday, we believe this works locally, right?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/response/model/SentenceAndOffenceDetails.kt
Gweaton: This is sorted now, we're only returning one field

src/main/resources/application-local.yml
anandelangovan1: yes, this works. However we've changed it to `hmpps-accredited-programmes-client-1`

Gweaton: ah can you amend this last commit and change it to that value?

anandelangovan1: Looks like we have already done it, the comment was from an initial commit its `hmpps-accredited-programmes-client-1` now

src/main/resources/application.yml
anandelangovan1: The tests read from application-test.yaml without these `services:` block the tests don't start up.
Application.yaml takes precedence. 

Does this answer your question?

src/main/resources/application-local.yml
Gweaton: I can still see this in this file, are you sure you're not looking at `application.yml` (rather than `application-local.yml`?

Gweaton (approved):

PR #162, opened 2023-11-14
#1087 Add & edit fields for data returned from caselist endpoint.

nsmj63-dxw:
## Context

[> see #1087 on the Accredited Programmed Trello Board.](https://trello.com/c/HeepPzY6/1087-add-submittedon-date-coursename-and-audience-to-dashboard-endpoints-m)

Several new requirements for the structure of a returned referral summary were introduced in this ticket, including adding mandatory fields from other tables: `submittedOn` from `referral`, `courseName` from `course`, and `audiences` from the many-to-many relationship in `course` and `audience` (we have a join table `course_audience` for this). This necessitated editing the existing `ReferralRepository` queries to retrieve the correct information, so it felt prudent in this case to define this more complex data set as a JPA projection in a separate Data Transfer Object (`ReferralSummaryProjection`). This also necessitated changing the transformation function `toApi()` to implement more complex business logic - offerings for a particular course may be on multiple streams (called `audiences` in the database), therefore any referral will be associated with (potentially) multiple streams, which results in multiple records for a single referral in the database (just with different streams), so it became necessary to flatten these results down so that a referral will contain a course with an offering that might have multiple streams (i.e. `List<Audience>`). By moving the transformer call into the service layer it becomes easier to test the "before" and "after"-transformation versions of a referral summary.

During implementation it also made sense to simplify the `referral` entity. Previously it only used the foreign key of its associated course offering (`offeringId`). To align more with the JPA specification, this has been changed to an entire `offering` object, which did not change underlying functionality but allows for greater scope of field-retrieval, whilst also aligning with other entity-relationship mapping already done via JPA in this codebase, e.g. between `offering` and `course`.

_NB: The above also necessitated updating the test data in our repeatable flyway migration `R__test_data.sql`, namely to add dummy information to the `audience` and `course_audience` tables. Separate but related; I noticed on my machine that previous flyway migrations which had been edited/combined for clarity, namely migrations 22, 23, and 24, now used different checksums which would fail at the flyway validation step of running the application. Running `flyway repair` does not work as our run configuration is not set up to require this. A fix for me was temporarily registering a flyway bean somewhere in a spring configuration class (doesn't matter where, as long as it's annotated with `@Configuration`) to manually run `flyway.repair()`. I have not left this in the codebase as a) it's quite unlikely to come up again and b) it's out-of-scope for this PR._

## Changes in this PR

- Update `ReferralEntity` to use JPA entity relationship between itself and `OfferingEntity`, instead of only `offeringId`
- Update repeatable flyway migration for test data for consistency
- Implement DTO `ReferralSummaryProjection` to capture complex details on referrals from the database
- Remove legacy `ReferralSummaryEntity` and update transformer function `toApi()` to process `List<ReferralSummaryProjection>` into a flattened-via-`audiences` `List<ReferralSummary> model.
- Move now-more-complex transformer function into the service layer at `ReferralService`.
- Update `ReferralController` to handle the updated manner of retrieving and processing referral summaries.
- Update unit and integration testing across the `ReferralSummary` endpoint.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/read/ReferralSummaryProjection.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/read/ReferralSummaryProjection.kt
@@ -1,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.read
+
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.create.ReferralEntity
+import java.time.LocalDateTime
+import java.util.*
+
+data class ReferralSummaryProjection(
+  val referralId: UUID,
+  val courseName: String,
+  val audience: String,
+  val status: ReferralEntity.ReferralStatus,
+  val submittedOn: LocalDateTime?,
+  val prisonNumber: String,
+)
--- a/src/test/resources/db/migration/R__test_data.sql
+++ b/src/test/resources/db/migration/R__test_data.sql
@@ -1,32 +1,51 @@
+DELETE from course_audience;
+DELETE from offering;
+DELETE from prerequisite;
+DELETE from audience;
+DELETE from course;
+DELETE from referral;
 
-
-INSERT INTO course(course_id, identifier, name, description, alternate_name)
-VALUES ('d3abc217-75ee-46e9-a010-368f30282367', 'LC', 'Lime Course', 'Explicabo exercitationem non asperiores corrupti accusamus quidem autem amet modi. Mollitia tenetur fugiat quo aperiam quasi error consectetur. Fugit neque rerum velit rem laboriosam. Atque nostrum quam aspernatur excepturi laborum harum officia eveniet porro.', 'LC'),
-       ('1811faa6-d568-4fc4-83ce-41118b902421', 'SC', 'Super Course', 'corona volva conturbo suscipit civis.', 'AC++'),
-       ('d3abc217-75ee-46e9-a010-368f30282368', 'RC', 'RAPID Course', 'corona volva conturbo suscipit civis.', NULL),
-       ('d3abc217-75ee-46e9-a010-368f30282369', 'CC', 'Custom Course', 'corona custom suscipit civis.', 'AC++'),
-       ('d3abc217-75ee-46e9-a010-368f30282370', 'TC', 'Temp Course', 'Temp custom suscipit civis.', 'AC++');
+INSERT INTO course(course_id, identifier, name, description, alternate_name, referable)
+VALUES ('d3abc217-75ee-46e9-a010-368f30282367', 'LC', 'Lime Course', 'Sample description', 'LC', true),
+       ('1811faa6-d568-4fc4-83ce-41118b902421', 'SC', 'Super Course', 'Sample description', 'AC++', true),
+       ('d3abc217-75ee-46e9-a010-368f30282368', 'RC', 'RAPID Course', 'Sample description', NULL, true),
+       ('d3abc217-75ee-46e9-a010-368f30282369', 'CC', 'Custom Course', 'Sample description', 'Custom', true),
+       ('d3abc217-75ee-46e9-a010-368f30282370', 'TC', 'Temp Course', 'Sample description', 'Temp', true);
 
 INSERT INTO course_participation (course_participation_id, prison_number, course_name, source, detail, location, type, outcome_status, year_started, year_completed, created_by_username, created_date_time, last_modified_by_username, last_modified_date_time)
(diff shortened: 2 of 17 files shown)


src/test/resources/db/migration/R__test_data.sql
```diff
@@ -1,32 +1,51 @@
+DELETE from course_audience;
```
Gweaton, line 1: Not quite sure I understand what this commit is doing from the commit message:

> Updated latest src/main (V24__) and src/test (R__) flyway migrations for consistency.

Could we add a bit more detail about what we mean by "for consistency"?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/read/ReferralSummaryProjection.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.read
+
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.create.ReferralEntity
+import java.time.LocalDateTime
+import java.util.*
+
+data class ReferralSummaryProjection(
+  val referralId: UUID,
+  val courseName: String,
+  val audience: String,
```
Gweaton, line 10: am a little confused on the use of `audience` here vs `audiences` elsewhere, particularly in terms of what the DTO looks like. Should the UI expect `audience` or `audiences`?

Gweaton (commented):
I think this broadly makes sense, just had a couple of questions

src/test/resources/db/migration/R__test_data.sql
Gweaton: Ah I've just read the NB of the PR description. Could we add that to the commit message please?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/read/ReferralSummaryProjection.kt
nsmj63-dxw: I've made this distinction because we're only ever expecting a single value inside the `audience` column of our database projection, but when transforming it to the api model structure that we send to the UI, we need to collate audiences together.

For instance, for our test data in `R__test_data.sql`, the repository will produce the following for `organisationId` "BWN":

```
| referralId                           | courseName   | audience                 | status           | submittedOn | prisonNumber |
|--------------------------------------|--------------|--------------------------|------------------|-------------|--------------|
| 0c46ed09-170b-4c0f-aee8-a24eeaeeddaa | Super Course | Gang offence             | REFERRAL_STARTED | null        | 1uXTfdH      |
| 0c46ed09-170b-4c0f-aee8-a24eeaeeddaa | Super Course | General violence offence | REFERRAL_STARTED | null        | 1uXTfdH      |
| 0c46ed09-170b-4c0f-aee8-a24eeaeeddaa | Super Course | Extremism offence        | REFERRAL_STARTED | null        | 1uXTfdH      |
```

JPA then builds out three `ReferralSummaryProjection` objects:
```
ReferralSummaryProjection(referralId=0c46ed09-170b-4c0f-aee8-a24eeaeeddaa, courseName=Super Course, audience=Gang offence, status=REFERRAL_STARTED, submittedOn=null, prisonNumber=1uXTfdH)
ReferralSummaryProjection(referralId=0c46ed09-170b-4c0f-aee8-a24eeaeeddaa, courseName=Super Course, audience=General violence offence, status=REFERRAL_STARTED, submittedOn=null, prisonNumber=1uXTfdH)
ReferralSummaryProjection(referralId=0c46ed09-170b-4c0f-aee8-a24eeaeeddaa, courseName=Super Course, audience=Extremism offence, status=REFERRAL_STARTED, submittedOn=null, prisonNumber=1uXTfdH)
```

The service layer calls the transformer method `toApi()` on this returned collection, which produces the equivalent `ReferralSummary` model for the UI to read:

```
ReferralSummary(id=0c46ed09-170b-4c0f-aee8-a24eeaeeddaa, courseName=Super Course, audiences=[Gang offence, General violence offence, Extremism offence], status=referralStarted, prisonNumber=1uXTfdH, submittedOn=null)
```

The controller then makes the list of `ReferralSummary` objects pageable, which I'll omit here for brevity.

So it made sense to have the projection only return a single `audience` and then have the actual model use `audiences`.

Gweaton: Thanks, that makes sense to me.

Gweaton (approved):

anandelangovan1 (approved):

PR #165, opened 2023-11-17
Fix seed data SQL

yndajas:
In a recent commit, I attempted to add an extra seeded referral, but I left a semi-colon at the end of the line adding the first seeded referral where there should now be a comma


Gweaton (approved):
To quote John, 🦅 👁️

PR #185, opened 2023-12-05
#1122 Implement new ReferrerUser entity and update Referrer accordingly.

nsmj63-dxw:
## Context

[> see #1122 on the Accredited Programmes Trello board.](https://trello.com/c/gulvnZLL/1122-add-new-user-model-and-link-to-referrals-to-replace-referrerid-l)

## Changes in this PR

A requirement for R&M is that a user can see specifically all referrals they have made, not just their organisation. Given this, we needed to represent users in our data structures. Currently we only use an API-generated referrerId when creating new referrals, which is no longer sufficient.

This PR introduces the usage of Spring Security's `SecurityContextHolder` to retrieve the username and role of the current logged-in user. This is used for creating a referral, and is persisted in its own table as `ReferrerUser`, named as such to allow scope for building other types of user (also, `user` is a protected keyword in all SQL dialects, so we needed some level of specificity). If a user already exists, a referral is simply assigned to them instead to avoid duplication. The referral table is now the source of truth for which users 'own' which referrals, and can be leveraged in later tickets relating specifically to reading user information.

- Add ReferrerUser entity and supporting table. Add new flyway migration and test information accordingly.
- Update Referrer to be the source of truth for which users own which referrals.
- Implement Spring Security to allow for username-based creation and reading of referrals.
- Add user-specific unit and integration testing for the new control flow.

Other smaller quality-of-life improvements have been made on the testing side to further assist this functional change, namely:
- cleaning up OAuth authentication to be homogenous across all testing that utilises it.
- moving duplicate testing constants to their own class.
- updating immediately-relevant tests which directly instantiate entities to instead use factory-based instantiation.
- updating the RepositoryTestBase to rely less on direct transaction management in favour of entity management.

## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/restapi/transformer/ReferralTransformersTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/restapi/transformer/ReferralTransformersTest.kt
@@ -8,6 +8,8 @@ import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.api.model.Refer
 import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.api.model.ReferralStatus.awaitingAssessment
 import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.api.model.ReferralStatus.referralStarted
 import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.api.model.ReferralStatus.referralSubmitted
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.common.util.CLIENT_USERNAME
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.common.util.PRISON_NUMBER
 import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.create.ReferralEntity.ReferralStatus.ASSESSMENT_STARTED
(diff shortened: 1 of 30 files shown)


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/restapi/transformer/ReferralTransformersTest.kt
```diff
@@ -23,10 +25,6 @@ import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.u
 
 class ReferralTransformersTest {
 
-  companion object {
```
Gweaton, line 26: Why has this gone?

Gweaton: ah looks like it's removed in a couple of commits' time

Gweaton (approved):
This looks good to me. Would be good for @anandelangovan1 to take a look too

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/restapi/transformer/ReferralTransformersTest.kt
nsmj63-dxw: Ah, must've put that change in the wrong commit. Apologies. Thought I'd caught them all.

PR #193, opened 2023-12-06
Ensure seeds exist for UI stubbed `activeCaseLoadId`

yndajas:
The UI stubs an active case load with ID `MRI`, so we need to ensure that there are (enough) referrals to offerings with this organisation ID. This does that, and also makes most of the courses referable

Related PRs:
- #191
- #189
- #188
- #187
- https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/392
- https://github.com/ministryofjustice/hmpps-accredited-programmes-ui/pull/391


Gweaton (approved):
Have tested this locally at that endpoint - 🤞 !

PR #245, opened 2024-01-19
#997 Implement EntityManager-based testing baseline in place of repeatable flyway migration

nsmj63-dxw:
## Context

> [see #997 on the Accredited Programmes Trello board.](https://trello.com/c/NWa5sSwg/997-fix-and-refactor-pact-tests-m)

The current setup of both our contract and integration testing framework require the use of a repeatable flyway migration, `R__test_data.sql`, which is implemented at context-level. The singular nature of this migration was questioned, as it contradicts modern testing conventions, such as simplicity, only instantiating the specific data one needs to test, and so on.

## Changes in this PR

Several approaches to this issue were attempted in this PR, nearly all of which were failures. This included:
1. **Factory-based entity instantiation (with `{Object}EntityFactory` classes)**, which aligns with our unit testing methodology. This fails because of our approach to generating our primary keys on new entities (our `id` fields). They are annotated with `@GeneratedValue`, which means that attempting to manually set them for testing purposes was overridden by JPA. This is an inherent 'feature' of that particular annotation, designed to prevent the direct updating of a primary key field at runtime; this makes sense in production code because one should have no reason to update an existing entity to change its identifier, but for our Pact-constrained contract testing it caused headaches.
2. **Multiple repeatable flyway migrations for specific entity relationships**. Whilst it is possible to implement multiple repeatable migrations, flyway is a bit funny with the way these migrations are applied unless you are _very_ specific with testing configuration. This felt like moving in the wrong direction, considering our goal to make contract testing easier to manage.
3. The successful option was **EntityManager-based direct SQL insertion for each entity-relationship**, which is both functionally-similar to the way that our repository-layer integration tests work, and allows for our service-layer integration tests to remain relatively unscathed. This involved creating a `PersistenceHelper` which contained the entity-creation methods, and implementation of it into the `IntegrationTestBase`, which is extended both across the contract tests and (as might be expected) across the integration tests. This allows for relatively easy editing of entities programatically, whilst ensuring that only the specific data required for individual tests is applied across the board, making for less brittle tests overall.

Additionally, several approaches to the issue of splitting out the singular `PactContractTest` class for easier management of each entity-relationship were attempted, also resulting in failures:
1. **Splitting the class into a `Base` and multiple extenders**, as is done with the integration tests. Pact configuration disallowed this from happening due to overlapping contexts being created simultaneously for each child class.
2. **Nested classes for each entity-relationship inside the `PactContractTest` class**, as is done with some unit tests. Again, pact configuration disallowed this from happening, this time due to being unable to find the tests within the nested classes.
3. The successful, but limited option was to **Stick to the status quo, feeding in specific data to each contract test manually**. Whilst this is not the most elegant solution, it does allow us to see, at a glance, which data each test requires to work correctly. As the number of tests we have at the moment is relatively small, this is a fair, if limiting, solution for now, though it is worth revisiting periodically particularly if further entity-relationships are required in the system.**
4. A potential solution for the future is to **Follow the strategy implemented in [hmpps-interventions-service](https://github.com/ministryofjustice/hmpps-interventions-service/blob/main/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsinterventionsservice/integration/pact/PactTest.kt#L17)**. The main concern with this approach is that it is quite complex; possibly overly-complex for our current needs. Nevertheless, it remains a potentially viable option - alongside some changes to the way we determine the current user - for later contract testing work.

## Release checklist

### Pre-merge

### Post-merge


src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/common/config/PersistenceHelper.kt
```diff
@@ -0,0 +1,170 @@
...
+    entityManager.createNativeQuery("DELETE FROM course_participation").executeUpdate()
+    entityManager.createNativeQuery("DELETE FROM referral").executeUpdate()
+    entityManager.createNativeQuery("DELETE FROM offering").executeUpdate()
+    entityManager.createNativeQuery("DELETE FROM course_audience").executeUpdate()
+    entityManager.createNativeQuery("DELETE FROM course").executeUpdate()
+    entityManager.createNativeQuery("DELETE FROM audience").executeUpdate()
+    entityManager.createNativeQuery("DELETE FROM referrer_user").executeUpdate()
+  }
+
+  fun setupCourses() {
```
Gweaton, line 27: Rather than creating everything up front and blowing it away between tests, how would you feel about extracting these into some helper like `createCourse(id:)` that inserts the specific entities with the ids we want, that could then be called in each `@State`? That way it would be really easy to tell exactly what entities we have lying about in each test state, and we wouldn't have to create so many things for each test?

I don't think this needs to block the PR, as this is definitely a fantastic step in the right direction, but having a smaller number of entities in each state will definitely help readability etc. This could be an enhancement of the current approach.

Gweaton (approved):
This is so much nicer than the previous approach, thanks for figuring it out. Have a suggestion for what I think would be a useful next step.

PR #704, opened 2024-11-11
APG-535 Add setup-service-pod script for easy service pod management

pfcairns:
## Changes in this PR

- Introduce a new script `setup-service-pod.bash` to automate the creation and management of service pods. This script sets up IRSA-based authentication and allows users to perform authorized AWS commands. 
- Updated the README with usage instructions and examples for interacting with AWS RDS.

### DIFF
--- a/script/kubernetes-scripts/setup-service-pod.bash
+++ b/script/kubernetes-scripts/setup-service-pod.bash
@@ -1,0 +1,52 @@
+#!/bin/bash
+namespace=hmpps-accredited-programmes-preprod
+
+hostname=$(hostname)
+# Read any named params
+while [ $# -gt 0 ]; do
+
+   if [[ $1 == *"--"* ]]; then
+        param="${1/--/}"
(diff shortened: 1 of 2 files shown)


script/kubernetes-scripts/setup-service-pod.bash
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash
+namespace=hmpps-accredited-programmes-preprod
```
anandelangovan1, line 2: Can namespace be passed as a param when executing the bash script?

anandelangovan1 (dismissed):

script/kubernetes-scripts/setup-service-pod.bash
pfcairns: Spot on - going to address this now. There is another bug in the script also, which I'm going to fix.

anandelangovan1 (approved):

PR #742, opened 2024-12-09
[APG-454] Check if staff details exisit in db before fetching

anandelangovan1:
## Release checklist

### Pre-merge

### Post-merge

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/StaffService.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/StaffService.kt
@@ -1,12 +1,12 @@
 package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.service
 
+import org.slf4j.LoggerFactory
 import org.springframework.stereotype.Service
 import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.client.nomisUserRoleManagementApi.model.StaffDetailResponse
 import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.common.exception.BusinessException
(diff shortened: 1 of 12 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/StaffService.kt
```diff
@@ -15,24 +15,37 @@ class StaffService(
...
-    staffDetailResponse?.staffId?.let {
-      staffRepository.findByStaffId(it)
-        ?: return staffRepository.save(buildStaffEntity(staffDetailResponse))
+  fun fetchPomDetailsIfNotAlreadyExists(staffId: Int?, prisonNumber: String, pomType: PomType): StaffEntity? {
+    if (staffId == null) {
+      log.warn("No $pomType pom staffId found for $prisonNumber")
+      return null
+    }
+
+    val findByStaffId = staffRepository.findByStaffId(staffId)
```
pfcairns, line 39: Naming : it's an odd name for a variable. Could you please rename.

pfcairns (commented):
Curious as to why the id has changed from BigInteger to Int? I would have thought for a key it'd be safer to use BigInteger.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/StaffService.kt
anandelangovan1: That's like the default suggestion for a variable in intellij. Updated and pushed

anandelangovan1:
> Curious as to why the id has changed from BigInteger to Int? I would have thought for a key it'd be safer to use BigInteger.

These are staffId which don't seem to be big enough for using BigInt and we are not doing any computations with it. I had BigInt initially when i dint know what to expect from moic api

PR #806, opened 2025-02-25
APG-711 Add ldc and ldc override flags

anandelangovan1:
## Release checklist

### Pre-merge

### Post-merge


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/create/ReferralEntity.kt
```diff
@@ -57,4 +57,6 @@ data class ReferralEntity(
   var overrideReason: String? = null,
   var transferReason: String? = null,
   var originalReferralId: UUID? = null,
+  var hasLdc: Boolean = false,
+  var hasLdcBeenOverwrittenByProgrammeTeam: Boolean = false,
```
pfcairns, line 61: Is it overwritten or overidden?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/OasysService.kt
```diff
@@ -307,6 +307,8 @@ class OasysService(
 
   fun getAlcoholDetail(assessmentId: Long): OasysAlcoholDetail? = fetchDetail(assessmentId, oasysApiClient::getAlcoholDetail, "AlcoholDetail")
 
+  fun getLDCScore() = (3..5).random().toBigDecimal()
```
pfcairns, line 310: Might be worth a comment or TODO to highlight that this is temporary

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/create/ReferralEntity.kt
```diff
@@ -57,4 +57,6 @@ data class ReferralEntity(
   var overrideReason: String? = null,
   var transferReason: String? = null,
   var originalReferralId: UUID? = null,
+  var hasLdc: Boolean = false,
```
pfcairns, line 60: Should this be  nullable and null by default? Can we be sure that Oasys will always return a score?

src/main/resources/db/migration/V120__add_ldc_and_ldc_override_flag_to_referral.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE referral
+    ADD COLUMN has_ldc BOOLEAN NOT NULL DEFAULT false;
```
pfcairns, line 2: Might be worth checking if this could be nullable, at least in our system. As a value of False has meaning in this context.

pfcairns (commented):
A few minor questions and comments.

src/main/resources/db/migration/V120__add_ldc_and_ldc_override_flag_to_referral.sql
anandelangovan1: good point, from the spec it seems like its a non nullable field

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/OasysService.kt
anandelangovan1: The new oasys endpoint will be in preprod this Friday mate, so this code will definitely change

src/main/resources/db/migration/V120__add_ldc_and_ldc_override_flag_to_referral.sql
thomaswilsonxyz: can we do some tests against the OASys API to just verify this ? if we've made the wrong assumption, we'll get a null-pointer exception and fail out, right ?

even if we don't fail-out and set to `FALSE` by default (as specified below) - are we okay with the chance or possibly setting someone as "does not have LDC" when in reality they do?  I genuinely don't know our appetite for this risk.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/create/ReferralEntity.kt
pfcairns: If a layer 3 assessment hasn't been done for a prisoner, then I'm guessing no score will be returned. Might be worth clarifying with the Capita team.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/OasysService.kt
anandelangovan1: Added a TODO comment

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/entity/create/ReferralEntity.kt
pfcairns: Overridden probably sounds best here.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/PniService.kt
```diff
@@ -200,6 +200,8 @@ class PniService(
     else -> null
   }
 
+  fun getLDC() = oasysService.getLDCScore() >= BigDecimal("2.99")
```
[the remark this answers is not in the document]
anandelangovan1: Updated. 
If the value is less than 3 its non ldc (false) and anything greater than 2.99 its true

src/main/resources/db/migration/V120__add_ldc_and_ldc_override_flag_to_referral.sql
pfcairns: The more I mull this over the more I think both fields should be nullable, as a null value here would denote lack of data, whereas a false would mean "does not have LDC" as stated above.

src/main/resources/db/migration/V120__add_ldc_and_ldc_overwritten_flag_to_referral.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE referral
+    ADD COLUMN has_ldc BOOLEAN NOT NULL DEFAULT false;
+
+ALTER TABLE referral
+    ADD COLUMN has_ldc_been_overwritten_by_programme_team BOOLEAN NOT NULL DEFAULT false;
```
KirilKolevMoJ, line 5: If we are going to stick with overridden in the ReferralEntity shouldn't this be consistent and have has_ldc_been_overridden_by_programme_team instead of has_ldc_been_overwritten_by_programme_team?

PR #812, opened 2025-02-26
[APG-730] Bump up hibernate version and associated refactor

pfcairns:
## Changes in this PR

- Upgraded Hibernate to version 6.6.9.Final which broke 100+ Integration tests as a result of stricter checks when persisting detached entities.
- This was fixed by:
      1.  Ensuring all Entities have a default value of null for their @Id fields
      2. Updating relevant tests and entities to reflect these changes.

### DIFF
--- a/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/domain/repository/CourseParticipationRepositoryTest.kt
+++ b/src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/domain/repository/CourseParticipationRepositoryTest.kt
@@ -45,6 +45,7 @@ class CourseParticipationRepositoryTest {
       .withYearCompleted(Year.parse("2022"))
       .produce()
     var courseParticipation = CourseParticipationEntityFactory()
+      .withId(null)
       .withCourseName("Course name")
       .withPrisonNumber(PRISON_NUMBER_1)
       .withSource("Source of information")
@@ -70,6 +71,7 @@ class CourseParticipationRepositoryTest {
   @Test
   fun `CourseParticipationRepository should save and retrieve CourseParticipationEntity objects, having all nullable fields set to null`() {
     var courseParticipation = CourseParticipationEntityFactory()
+      .withId(null)
       .withCourseName(null)
       .withPrisonNumber(PRISON_NUMBER_1)
       .withSource(null)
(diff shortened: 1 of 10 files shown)


KirilKolevMoJ (approved):

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/domain/repository/CourseParticipationRepositoryTest.kt
```diff
@@ -70,6 +71,7 @@ class CourseParticipationRepositoryTest {
   @Test
   fun `CourseParticipationRepository should save and retrieve CourseParticipationEntity objects, having all nullable fields set to null`() {
     var courseParticipation = CourseParticipationEntityFactory()
+      .withId(null)
```
anandelangovan1, line 74: Does this have to null or can we not set it at all?

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/domain/repository/CourseParticipationRepositoryTest.kt
```diff
@@ -107,6 +109,7 @@ class CourseParticipationRepositoryTest {
       .withYearCompleted(Year.parse("2022"))
       .produce()
     var courseParticipation = CourseParticipationEntityFactory()
+      .withId(null)
```
anandelangovan1, line 112: Similar to above comment

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/domain/repository/CourseParticipationRepositoryTest.kt
pfcairns: All our factory implementations seem to set the Id field to a random UUID which is useful elsewhere, just not here. If we don't explicitly set it, the test fails.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/unit/domain/repository/CourseParticipationRepositoryTest.kt
pfcairns: Same as above.

PR #873, opened 2025-04-15
APG-666 Remove usage of manageUsersApi 

anandelangovan1:
## Release checklist

### Pre-merge

### Post-merge


Giruvagen (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/NomisUserRolesService.kt
```diff
@@ -34,6 +35,26 @@ class NomisUserRolesService(val nomisUserRolesApiClient: NomisUserRolesApiClient
         AuthorisableActionResult.Success(null)
       }
     }
+    return staffDetail.entity
```
thomaswilsonxyz, line 38: for my kotlin knowledge - what is this doing now that wasn't previously being run ?

my understanding is that this statement will run whenever one of the `when` cases matches, but those look like they have implicit returns (or explicit `throw`s)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/NomisUserRolesService.kt
```diff
@@ -34,6 +35,26 @@ class NomisUserRolesService(val nomisUserRolesApiClient: NomisUserRolesApiClient
...
+  }
+
+  fun getUserDetail(username: String): UserDetail? {
+    val userDetail = when (val response = nomisUserRolesApiClient.getUserDetail(username)) {
```
thomaswilsonxyz, line 42: this looks repeated to the block above, is it worth extracting it? or would that be too DRY ?

thomaswilsonxyz (approved):
approved with some comments that might be more about my ignorance than your code :)

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/CaseNotesApiService.kt
```diff
@@ -160,7 +160,8 @@ class CaseNotesApiService(
   fun getFullName(): String? {
     val username = SecurityContextHolder.getContext().authentication?.name.orEmpty()
     return try {
-      manageUsersService.getUserDetail(username)?.name ?: username
+      val userDetail = nomisUserRolesService.getUserDetail(username)
+      userDetail?.fullName() ?: username
```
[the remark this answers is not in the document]
anandelangovan1: I thought the fallback username might be better, if it defaulted to username atleast we have a way to finding who it is in future if we need to. 
Also IMHO, "This message was created by MY_USERNAME" reads better compared to "This message was created by null"

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/NomisUserRolesService.kt
anandelangovan1: Good question. I did this in 

https://github.com/ministryofjustice/hmpps-accredited-programmes-api/blob/804ad85ae50f999634b7ca4fc32c534ba0d70634/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/client/oasysApi/OasysApiClient.kt

This class only has 2 methods, so might be a bit premature. Could look at other classes for have a pattern for sure

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/service/NomisUserRolesService.kt
anandelangovan1: When `staffDetail.entity` runs only when the call is successful. Other cases we throw an error / null

PR #889, opened 2025-04-29
APG-655: Fix the updateReferralById method

thomaswilsonxyz:
**WHAT**

Update the `ReferralService:updateReferralById` method to only update
the `additionalInformation` and `referrerOverrideReason` fields if the
incoming object has a field.

**WHY**

The `UpdateReferral` object is intentionally a partial, so we know that
the fields are nullable strings, but the code is treating them as though
they will always be present.


anandelangovan1:
I think we agreed on the full referral being passed when an update is made, similar to the below:

snippet from UI codebase:

```
    this.referralService.updateReferral(username, referralId, {
          ...referral,
          hasLdc: !referral.hasLdc,
          hasLdcBeenOverriddenByProgrammeTeam: true,
        }),
      ])

```

thomaswilsonxyz:
> I think we agreed on the full referral being passed when an update is made, similar to the below:
> 
> snippet from UI codebase:
> 
> ```
>     this.referralService.updateReferral(username, referralId, {
>           ...referral,
>           hasLdc: !referral.hasLdc,
>           hasLdcBeenOverriddenByProgrammeTeam: true,
>         }),
>       ])
> ```

We may have done, but the code is annotated and typed as having it as optional, and our FE also treats the fields as though they're optional - I think it's the smallest, most pragmatic change we can make - as opposed to updating code in multiple repos.

anandelangovan1 (approved):

PR #910, opened 2025-05-14
Gardening: Improve the SexualOffenceDetails model

thomaswilsonxyz:
- **refactor(StaffServiceTest.kt): Don't use inline variable in a test expectation, to make the test more resilient**
- **fix(ReferenceDataController): Remove an errenous 4xx response code**
- **fix(SexualOffenceDetails): Update documentation and types**

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/model/SexualOffenceDetails.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/model/SexualOffenceDetails.kt
@@ -2,19 +2,25 @@ package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.restapi.model
 
 import com.fasterxml.jackson.annotation.JsonProperty
 import io.swagger.v3.oas.annotations.media.Schema
+import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.referencedata.type.SexualOffenceCategoryType
 import java.util.*
 
 data class SexualOffenceDetails(
   @Schema(example = "67ea1478-e4c3-46be-8b7f-86833fb87540", description = "The unique sexual offence identifier")
   @get:JsonProperty("id") val id: UUID,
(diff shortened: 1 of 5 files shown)


pfcairns (commented):
Just the one minor question.

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/model/SexualOffenceDetails.kt
```diff
@@ -2,19 +2,25 @@ package uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.restapi.model
...
+  @Schema(example = "Evidence of ritualism in the offence", description = "A human-friendly description of the Offence itself")
   @get:JsonProperty("description") val description: String,
-  @Schema(example = "Any use of additional violence or force not necessary to gain victim compliance", description = "Further elaborations of the offence")
+
+  @Schema(example = "Fixed actions or words, performed in a specific way", description = "Further elaborations of the offence")
   @get:JsonProperty("hintText") val hintText: String? = null,
-  @Schema(example = "1", description = "The score associated with the offence")
+
+  @Schema(example = "INCLUDES_VIOLENCE_FORCE_HUMILIATION", description = "A computer-friendly identifier for the Offence's category")
+  @get:JsonProperty("categoryCode") val categoryCode: SexualOffenceCategoryType,
```
pfcairns, line 19: Just a thought : if we're returning the enum in its entirety we can do away with the separate `categoryCode` and `categoryDescription` fields no?

thomaswilsonxyz: Good question ! I think we already have this distinction with the `SexualOffenceDetailsEntity` vs `SexualOffenceDetails` - we still want to maintain both fields them coming over the API, right, so the front-end can see and use them.

I don't see much benefit here in updating the response payload to e.g. 

```json
{ 
  "category": {
    "value": "VALUE",
    "description": "value description"
  }
}
```

so I _think_  we're okay to leave this as it is here

pfcairns (approved):

PR #1051, opened 2026-04-02
Update branch name validation in GitHub Action 

pfcairns:
## Changes in this PR

- Updated branch name validation in GitHub Actions to prevent use of external GHA


Giruvagen (approved):

Giruvagen (changes requested):
needs to be changed to your Jira format in the regex

Giruvagen (approved):

PR #1052, opened 2026-04-02
Remove `git.yml` workflow for blocking fixup and squash merges

pfcairns:
## Changes in this PR

- Removed the `git.yml` workflow that previously prevented fixup and squash merge commits as it contains prohibited action.


niaevans9 (approved):
lgtm

PR #1068, opened 2026-06-03
Fix: Add @JsonIgnoreProperties to AlertCodeSummary to handle upstream API changes

raby:
**What**
Added @JsonIgnoreProperties(ignoreUnknown = true) to the AlertCodeSummary data class, matching the pattern already used on the sibling Alert model.

**Why**
The upstream Prisoner Alerts API added a new canBeAdministered field to the alertCode object. Our model didn't tolerate unknown properties, causing Jackson UnrecognizedPropertyException on every call to retrieve alerts. This meant OasysService.getActiveAlerts() returned null for all prisoners — alert data was completely missing from referral detail pages.

**Prod log evidence:**
WARN OasysService : Failure to retrieve ActiveAlerts for prisonNumber A3744AJ reason
com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "canBeAdministered"

**Changes**
AlertCodeSummary.kt — Added @JsonIgnoreProperties(ignoreUnknown = true) annotation
Wiremock fixtures — Added canBeAdministered field to test JSON to prove the fix works and prevent regression

**Testing**
PrisonerAlertsApiClientIntegrationTest passes with the new field present in fixtures
The test now exercises the exact scenario that was failing in production


pfcairns (approved):
LGTM

PR #1087, opened 2026-07-16
APG-2492: resolve staff surnames for SAR custody report

raby:
Introduce StaffLookupService (thin wrapper over StaffRepository) exposing findLastNameByUsername(String?) and findLastNameByStaffId(BigInteger?), both returning null when no matching staff row exists. Keeping this as a service (rather than calling the repository directly) gives us a single seam for future caching / batching / fallback-rule tweaks.

Wire the lookup into SubjectAccessRequestService at 8 sites:

  - SarAuditRecord.referrerUsername            (username -> surname)
  - SarAuditRecord.auditUsername               (username -> surname, widened String -> String?)
  - SarCourseParticipation.createdByUser       (username -> surname)
  - SarCourseParticipation.updatedByUser       (username -> surname)
  - SarReferral.referrerUsername               (username -> surname)
  - SarReferral.primaryPomStaffId              (BigInteger? -> primaryPomStaffSurname: String?)
  - SarReferral.secondaryPomStaffId            (BigInteger? -> secondaryPomStaffSurname: String?)
  - SarReferralStatusHistoryEntity.username    (username -> surname, widened String -> String?)

Null values render as "No Data Held" via the existing optionalValue mustache helper (agreed with QAT per section 2.2 of docs/2026-07-14-SAR-custody-QAT-round2-review-notes.md in the manage-and-deliver-api repo).

Also drop the now-redundant getUserLastName template helper calls in sar_template.mustache -- the service already yields surnames, so the template just needs optionalValue -- and rename the two POM label/field placeholders accordingly.

Test updates:
  - New StaffLookupServiceTest covering happy path, no-match, null and blank inputs.
  - SubjectAccessRequestServiceTest stubs the new service and asserts surname values.
  - SarContractIntegrationTest seeds a staff row keyed on TEST_USER so username / staffId lookups exercise the resolved-surname path; SAR contract fixtures updated to match.
  - SubjectAccessRequestServiceIntegrationTest seeds a staff row and asserts surname resolution end-to-end.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
```diff
@@ -12,6 +12,12 @@ interface StaffRepository : JpaRepository<StaffEntity, UUID> {
 
   fun findByStaffId(staffId: BigInteger): StaffEntity?
 
+  @Query("SELECT s.lastName FROM StaffEntity s WHERE s.username = :username")
+  fun findLastNameByUsername(username: String): String?
```
miroslav25a: The staff.username is not uniques, so I guess this method should return a list of strings?

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
```diff
@@ -12,6 +12,12 @@ interface StaffRepository : JpaRepository<StaffEntity, UUID> {
...
+
+  @Query("SELECT s.lastName FROM StaffEntity s WHERE s.staffId = :staffId")
+  fun findLastNameByStaffId(staffId: BigInteger): String?
```
miroslav25a: The staff.staffId is not uniques, so I guess this method should return a list of BigInteger?

raby: Good spot I have addressed as a commit

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
raby: Good spot I have addressed as a commit

miroslav25a (dismissed):

miroslav25a (approved):

PR #1090, opened 2026-07-17
APG-2342 Make `bookingId` and `prisonerNumber` nullable in response

pfcairns:
## Changes in this PR

- Updated the `PeopleSearchResponse` model to allow `bookingId` and `prisonerNumber` to be nullable.

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/model/PeopleSearchResponse.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/model/PeopleSearchResponse.kt
@@ -26,10 +26,10 @@ import io.swagger.v3.oas.annotations.media.Schema
 data class PeopleSearchResponse(
 
   @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("bookingId", required = true) val bookingId: String,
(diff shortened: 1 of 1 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/model/PeopleSearchResponse.kt
```diff
@@ -26,10 +26,10 @@ import io.swagger.v3.oas.annotations.media.Schema
 data class PeopleSearchResponse(
 
   @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("bookingId", required = true) val bookingId: String,
+  @get:JsonProperty("bookingId", required = true) val bookingId: String? = null,
 
   @Schema(example = "null", required = true, description = "")
-  @get:JsonProperty("prisonerNumber", required = true) val prisonerNumber: String,
+  @get:JsonProperty("prisonerNumber", required = true) val prisonerNumber: String? = null,
```
Jack-Dempster: should these be `required = true` if they are able to be null? And are there any better examples you could add to the schema?

pfcairns: Yep, totally fair will address.

Jack-Dempster (approved):
👍

PR #1096, opened 2026-07-20
APG-2342 Add admin endpoint and documentation to handle NOMIS prisoner number merges.

pfcairns:
## Changes in this PR

- Added admin endpoint that updates a prisoner number to be used on an adhoc basis
- Added integration tests to verify behaviour
- Create markdown file detailing NOMIS prisoner merged events and how to use the API endpoints.


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/restapi/controller/AdminController.kt
```diff
@@ -110,7 +124,66 @@ class AdminController(
...
+    @RequestBody prisonerNumberUpdateRequest: PrisonerNumberUpdateRequest,
+  ): ResponseEntity<String> {
+    // verify that the current prisoner number exists within AcP
+    val person = personRepository.findPersonEntityByPrisonNumber(prisonerNumberUpdateRequest.currentPrisonerNumber)
+      ?: log.error("Prisoner number: ${prisonerNumberUpdateRequest.currentPrisonerNumber} not found in ACP").also {
+        throw NotFoundException("Prisoner number: ${prisonerNumberUpdateRequest.currentPrisonerNumber} not found in ACP")
+      }
+    // verify that the new prisoner number exists in NOMIS
+    val personSearchResponse = peopleSearchApiService.searchPeople(PeopleSearchRequest(prisonerIdentifier = prisonerNumberUpdateRequest.newPrisonerNumber))
+    if (personSearchResponse.isEmpty() || !prisonerNamesMatch(person as PersonEntity, personSearchResponse[0])) {
```
raby: The if condition combines two distinct failure scenarios into a single block, both throwing NotFoundException (HTTP 404)

When personSearchResponse.isEmpty() → correctly a 404

When !prisonerNamesMatch(...) → should be a 400 Bad Request (prisoner exists in NOMIS but the name doesn't match). Your own Swagger doc on this endpoint even documents it as a 400 with description "prisoner names do not match between systems", but the code always throws a 404.

We could split into two separate checks. The name-mismatch branch should throw BusinessException (which maps to 400 via ApiExceptionHandler):

```
if (personSearchResponse.isEmpty()) {
  log.error("New prisoner number ${prisonerNumberUpdateRequest.newPrisonerNumber} not found in NOMIS")
  throw NotFoundException("New prisoner number: ${prisonerNumberUpdateRequest.newPrisonerNumber} not found in NOMIS")
}
if (!prisonerNamesMatch(person, personSearchResponse[0])) {
  log.error("Prisoner names do not match for new number ${prisonerNumberUpdateRequest.newPrisonerNumber}")
  throw BusinessException("Prisoner names do not match between ACP and NOMIS for prisoner number ${prisonerNumberUpdateRequest.newPrisonerNumber}")
}
```

pfcairns: Thanks Raby, good spot. Will address.

raby (approved):
Looks great

PR #1098, opened 2026-07-21
APG-2492: batch staff-surname lookups in SAR generation

raby:
Follow-up to the surname-lookup helper and V144 index migration.

Collapse the O(rows) staff-surname point-lookup pattern in SubjectAccessRequestService into exactly two batch queries per SAR request (one by-username, one by-staff-id).

StaffRepository gains findSurnamesByUsernames / findSurnamesByStaffIds returning lightweight projection interfaces (uses the V144 indexes).

StaffLookupService adds resolveSurnamesByUsername / resolveSurnamesByStaffId, which filter null/blank inputs, dedupe multi-row matches (logging a WARN so duplicates stay visible in prod), and return maps keyed by identifier.

SubjectAccessRequestService.getPrisonContentFor now collects all usernames / staff IDs referenced by the referral, participation, audit and status-history entity sets up-front, resolves both batch maps once, and threads a StaffSurnames context object into each mapper. Scalar findLastNameByUsername / findLastNameByStaffId helpers remain for any non-batch caller.

Adds unit tests for the new batch resolvers (dedupe, null/blank filter, empty short-circuit, unmatched keys) and integration tests for the new projections

### DIFF
--- a/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
+++ b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
@@ -7,6 +7,24 @@ import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.c
 import java.math.BigInteger
 import java.util.UUID
 
+/**
+ * Projection returned by [StaffRepository.findSurnamesByUsernames] so callers can
+ * build a `username -> lastName` map without loading full staff rows.
+ */
+interface UsernameSurnameProjection {
+  val username: String
+  val lastName: String
+}
+
+/**
+ * Projection returned by [StaffRepository.findSurnamesByStaffIds] so callers can
+ * build a `staffId -> lastName` map without loading full staff rows.
+ */
+interface StaffIdSurnameProjection {
+  val staffId: BigInteger
+  val lastName: String
+}
+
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
```diff
@@ -7,17 +7,73 @@ import uk.gov.justice.digital.hmpps.hmppsaccreditedprogrammesapi.domain.entity.c
...
+    ORDER BY s.username, s.id
+    """,
+  )
+  fun findSurnamesByUsernames(usernames: Collection<String>): List<UsernameSurnameProjection>
+
+  /**
+   * Batch equivalent of [findLastNameByStaffId]: returns one row per matched
+   * staff record so callers can group into a `staffId -> lastName` map in a
+   * single query. Uses idx_staff_staff_id (see V144).
+   *
```
Jack-Dempster, line 62: think this comment isn't relevant to this method

Jack-Dempster (approved):

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/StaffRepository.kt
raby: Thanks I shall reword

PR #1101, opened 2026-07-23
APG-2493: enrich SAR referrals with a nested originalReferral block

raby:
## What

Each `SarReferral` in the custody SAR payload currently carries a bare
`originalReferralId: UUID?` — a subject reading their SAR can't make any
sense of that. This PR adds a sibling `originalReferral: SarOriginalReferral?`
block with 9 human-readable fields, rendered as a nested "Original referral"
table under each referral in the SAR HTML report.

Fields (all sourced from the referenced original referral):

`id`, `courseName`, `organisationName`, `submittedOn`, `statusCode`,
`referrerSurname`, `referrerOverrideReason`, `hasLdc`, `additionalInformation`

The raw `originalReferralId` is kept alongside the nested block for
back-compat and debugging.

Ticket: [APG-2493](https://dsdmoj.atlassian.net/browse/APG-2493)
Sign-off: OSAR via Naseem (email 2026-07-2x) — all 9 fields confirmed.

## Why now (parallel with APG-2495)

APG-2493 was originally sequenced after APG-2495 (the post-deploy retest
of the live-like SAR pipeline). The regression-critical checks on APG-2495
(A1, B1, B3, C1, D1, E2) are already green on `main`; only the auth-gated
live checks remain, blocked on ops ticket HAAR-5793. Sharon needs a dev
report over to Offender SAR by end of Fri 2026-07-24 for the 2026-08-03
deadline, so this PR is being raised in parallel to add reviewable content.

## How — two commits, review each independently

**Commit 1 — `APG-2493: batch-load organisations by code`** (pure perf, no
behaviour change)
- New `OrganisationRepository.findAllByCodeIn(codes: Collection<String>)`
  (Spring Data derived — `WHERE code IN (?)`).
- Refactors `Content.organisations` construction in
  `SubjectAccessRequestService` from an N-call `findOrganisationEntityByCode`
  loop to a single batched call. Order + content preserved bit-exact.
- New `OrganisationRepositoryIntegrationTest` (matched / unmatched-dropped
  / empty).
- Existing `SubjectAccessRequestServiceTest` mock updated for the new
  method signature.

**Commit 2 — `APG-2493: enrich originalReferralId with SarOriginalReferral`**
- Adds `SarOriginalReferral` DTO + `originalReferral` field on `SarReferral`.
- Batch-loads originals via `referralRepository.findAllById(...)`.
- Widens the batch org lookup from commit 1 to include original-referral
  org codes; feeds `organisationsByCode` to the mapper.
- Folds `originalsById.values` into the existing `resolveStaffSurnames`
  collector so original-referral referrer surnames come for free from the
  same two batched staff queries (introduced by APG-2492).
- WARN log when an `originalReferralId` doesn't resolve (referential-
  integrity drift), mirroring the duplicate-staff-row WARN from APG-2492.
- Mustache template gains a `{{#originalReferral}}...{{/originalReferral}}`
  nested table, gated so it silently disappears when the block is null.
- Existing unit test extended in place to cover populated / orphan-
  unresolved / null-id branches on a single 3-referral SAR.
- SAR contract JSON fixture gains `"originalReferral":null` on the one
  seeded referral (which has no `originalReferralId`). HTML render fixture
  unchanged (nested block renders nothing when null).

## Query strategy — honest accounting

- **+1 batched query** for originals (`findAllById`).
- Organisation lookup is the **same batched call from commit 1**, now
  fed a union of codes across current + original referrals — still one
  query.
- Original-referral referrer surnames — **zero extra queries** (piggyback
  the existing `resolveStaffSurnames` collector).

**Caveat:** `findAllById` returns each original with `offering` and
`referrer` marked `FetchType.LAZY`, so the enrichment mapper triggers one
hydration select per unique offering + one per unique referrer (`course`
is EAGER on `OfferingEntity` so it piggybacks). Net cost is therefore
**`+1 batched + 2·(unique-originals)`** per SAR, not `+1` flat. Real-world
N is typically 0–1 so absolute impact is negligible — but flagged for
transparency and pre-planned as a JOIN FETCH follow-up (see below).

## Verification

```bash
./gradlew compileKotlin compileTestKotlin \
          ktlintMainSourceSetCheck ktlintTestSourceSetCheck \
          test --tests '*.SubjectAccessRequestService*' \
               --tests '*.SarContractIntegrationTest' \
               --tests '*.OrganisationRepository*' -q

[APG-2493]: https://dsdmoj.atlassian.net/browse/APG-2493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### CODE REVIEW

File: src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/OrganisationRepository.kt [hunk 1]
```diff
@@ -11,4 +11,14 @@ interface OrganisationRepository : JpaRepository<OrganisationEntity, UUID> {
...
   fun findOrganisationEntityByName(name: String): OrganisationEntity?
+
+  /**
+   * Batch-load organisations by their business code. Preferred over calling
+   * [findOrganisationEntityByCode] in a loop when resolving multiple codes
+   * (e.g. building the SAR content payload) as it collapses N point-lookups
+   * into a single `WHERE code IN (?)` query. Unmatched codes are silently
+   * dropped by the IN clause; callers should treat a missing entry in the
+   * `associateBy { it.code }` map as "code not found".
+   */
```
pfcairns: Minor: the JPA Repository method name is sufficient here. The comment isn't necessary.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/OrganisationRepositoryIntegrationTest.kt
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Integration coverage for the batch [OrganisationRepository.findAllByCodeIn]
+ * derived query. Guards the Spring Data method-name derivation against silent
+ * breakage from a future rename of `OrganisationEntity.code`, and pins the
+ * "unmatched codes are silently dropped" semantic the SAR content builder
+ * relies on.
+ */
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
+@ActiveProfiles("test")
```
pfcairns: The `ActiveProfiles` annotation is unnecessary here as it's present on the `IntegrationTestBase` superclass

pfcairns (commented):
Looks good. Just a couple of minor comments.

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/OrganisationRepositoryIntegrationTest.kt
raby: Great spot

src/main/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/OrganisationRepository.kt
raby: I have removed the comment

src/test/kotlin/uk/gov/justice/digital/hmpps/hmppsaccreditedprogrammesapi/domain/repository/OrganisationRepositoryIntegrationTest.kt
raby: I have removed

pfcairns (approved):
LGTM

PR #1102, opened 2026-07-23
APG-2510: remove staff.username from SAR staff section

raby:
## What

Drops the `username` field from the SAR `staff[]` payload so each staff
entry surfaces only `lastName`.

## Why

OSAR (via Naseem, 2026-07-2x) confirmed the custody SAR should surface
only staff **surnames** — usernames are internal identifiers that add
no value to the subject reading their SAR.

This closes the loop on the wider staff-attribute cleanup:

- **APG-2499** dropped `staffId`, `firstName`, `primaryEmail`,
  `accountType` from `SarStaff`.
- **APG-2492** replaced raw usernames / POM `staffId` on referral,
  participation, audit and status-history rows with resolved surnames
  via `StaffLookupService`.
- **APG-2510** (this PR) removes the last remaining `username` leak —
  the one on `SarStaff` itself.

## Changes

- `SubjectAccessRequestService.SarStaff` — drop `username`, retain
  `lastName`.
- `SubjectAccessRequestService.toSarStaff()` — drop the `username`
  mapping.
- Move `distinctBy { it.username }` from **after** the mapping to
  **before** it, so dedup-by-real-identity is preserved now that the
  DTO no longer carries a username. Two staff members who happen to
  share a surname still appear as two rows.
- `src/test/resources/sar/sar-api-response.json` — drop `"username"`
  from the `staff[0]` fixture element.

## Non-changes

- **No DB migration.** `staff.username` column stays — still used
  internally by `findByPrisonNumber`, batch surname lookups and
  `StaffLookupService`.
- **No `StaffLookupService` or repository changes.**
- **No mustache template change.** The `{{#staff}}` block already
  emitted only `lastName`; there was never a `Username` row rendered
  for the staff section.
- **No HTML render fixture change** (follows from the template being
  unchanged).

## Contract impact

The `staff[]` JSON contract changes from `{ lastName, username }` to
`{ lastName }`. The SAR aggregator only renders whatever fields are
present, and no internal consumer reads `.username` off SarStaff
(verified by workspace grep). The `SarContractIntegrationTest`
fixture diff is the only behavioural signal.

## Verification


niaevans9 (approved):
lgtm

PR #1107, opened 2026-08-03
APG-2546: remove auditRecords section from SAR

raby:
Removes the auditRecords section from the SAR payload per OSAR
round-2 review (Roxanne, DD spreadsheet rows 22-31).

Motivation: on retest PRN A8610DY the audit section contributed
28,483 rows to the SAR - the majority of the "8,000-page PDF"
complaint. Removing this section is the single largest OSAR
readability win in APG-2546.

Changes:
- Delete auditRecords from Content data class + Content(...) call
- Delete SarAuditRecord DTO + toSarAudit mapper
- Drop AuditRepository from SubjectAccessRequestService constructor
- Delete AuditRepository.getSarAuditRecords (dead after this PR)
- Delete the `<h2>Audit records</h2>` block from sar_template.mustache
- Delete audit-record seed, mock, and assertions from tests
- Regenerate SAR contract snapshots

Not changed:
- AuditEntity (still used by audit writers)
- Other AuditRepository methods (untouched)
- resolveStaffSurnames B1 test - call count is unaffected (audit
  rows only added usernames to already-batched sets; removing them
  does not change the two-query shape of the resolver)

Rollback: single git revert. Fixtures come back with the revert
commit; no schema state to rewind.

---

Notes for reviewer, beyond the doc's template:
- Also removed the equivalent audit seed + assertions in
  `SubjectAccessRequestServiceIntegrationTest.kt` (which the working
  doc did not list) since it was a compile-blocker after the
  `Content.auditRecords` field went away. The change mirrors the
  seed/assert removals in `SarContractIntegrationTest` and the unit
  test.
- Entity-schema snapshot verified unchanged (byte-identical to the
  regenerated `.log`), as expected per the doc.
- Full `./gradlew ktlintCheck test` passes locally (678 tests).


pfcairns (approved):
LGTM

PR #1109, opened 2026-08-03
APG-2546: remove referralStatusHistory + referralStatusReasons sections from SAR

raby:
APG-2546: remove referralStatusHistory + referralStatusReasons sections from SAR

Removes two coupled sections per OSAR round-2 review (Roxanne DD
spreadsheet rows 192–201 and 205–209 — all fields flagged red).

Coupling rationale: referralStatusReasons is derived in-service from
referralStatusHistory, so removing them together avoids an
intermediate state where we query history just to discard the
derived value.

Changes:
- Delete both fields from Content data class + Content(...) call
- Delete SarReferralStatusHistoryEntity + SarReferralStatusReason DTOs
- Delete the two mappers
- Drop ReferralStatusHistoryRepository from SubjectAccessRequestService
- Delete/keep ReferralStatusHistoryRepository.findByPrisonNumber
  based on grep of remaining callers (see PR body for decision)
- Delete both <h2> blocks from sar_template.mustache
- Delete seed, mock, and assertions from tests
- Regenerate SAR contract snapshots

resolveStaffSurnames is the second parameter shrink in APG-2546
(PR-1 removed auditRecords). Query count per SAR is unchanged.

## `findByPrisonNumber` decision

Deleted. Confirmed via `grep -rn "findByPrisonNumber" src/main` that
`SubjectAccessRequestService` was the sole `src/main` caller of
`ReferralStatusHistoryRepository.findByPrisonNumber`. Other services
(`ReferralStatusHistoryService`) use
`getAllByReferralIdOrderByStatusStartDateDesc` instead. Removing the
method here keeps the repository surface honest.

## Artefacts

| Item | Value |
|---|---|
| Sample SAR PDF pages (contract test seed data) | 3 |
| Snapshot files regenerated | `sar-api-response.json`, `sar-expected-render-result.html` |
| `entity-schema.json` | unchanged (deleting reads, not schema) |

## Verification

```
grep -rn "referralStatusHistory\|referralStatusReasons" src/main   # zero SAR-facing hits
grep -rn "SarReferralStatusHistoryEntity\|SarReferralStatusReason" src/main   # zero
grep -rn "toSarReferralStatusHistory\|toSarReferralStatusReason" src/main   # zero
grep -rn "findByPrisonNumber" src/main   # no ReferralStatusHistoryRepository hits
./gradlew ktlintCheck test                                          # green (678 tests)
```


pfcairns (approved):
LGTM

PR #1112, opened 2026-08-05
APG-2546: strip internal ID fields from remaining SAR sections

raby:
APG-2546: strip internal ID fields from remaining SAR sections

Removes `SarPerson.id`, `SarOrganisation.id`, and `SarReferral.originalReferralId` from the SAR payload per Roxanne's DD spreadsheet rows 105, 111 and 165 ("internal ref — should be a no" / "do not add the uuid"). Roxanne confirmed the `originalReferralId` fold-in in person 2026-08-04 pm (see `doc/planning/APG-2546/DELIVERY-LOG.md` "Roxanne in-person answers 2026-08-04 pm"). Also updates the person block of the mustache template to drop the corresponding row, and drops the top-level "Original referral ID" row from the referrals block — the nested `{{#originalReferral}}` block still renders "Original referral ID" from `SarOriginalReferral.id`, which is retained (the subject already sees the resolved original-referral sub-block, so that identifier is meaningful in context).

## Changes

- Delete `originalReferralId: UUID?` from `SarReferral` + positional argument in `toSarReferral` mapper (`SubjectAccessRequestService.kt`)
- Delete `id: UUID?` from `SarPerson` + `toSarPerson` mapper
- Delete `id: String` from `SarOrganisation` + `toSarOrganisation` mapper
- Delete "Original referral ID" row (line 14) and "Person ID" row from `sar_template.mustache`
- Drop three now-irrelevant assertions from `SubjectAccessRequestServiceTest` (`referral.originalReferralId` / `orphanReferral.originalReferralId` / `plainReferral.originalReferralId`) — the observable contract is now the nested `originalReferral` block, still asserted
- Regenerate SAR contract snapshots (`sar-api-response.json` + `sar-expected-render-result.html`)

## Not changed

- Entity-level `ReferralEntity.originalReferralId` — still required for the batch lookup that populates the `SarOriginalReferral` sub-block
- `SarOriginalReferral.id` / template "Original referral ID" row inside the `{{#originalReferral}}` block — the subject sees the resolved original-referral sub-block, so this identifier stays
- `{{#organisations}}` template block — never rendered an Id row in the first place; no code drift from Roxanne's expectation there, the row is already absent
- `SarPniResult` DTO — Roxanne's rows 127/128/131 flag `pniResultId` / `referralId` / `oasysAssessmentId` as internal, but those fields are already absent from the DTO. No action needed; planning doc §B has the audit trail
- Test files other than `SubjectAccessRequestServiceTest` — no other assertion probes the two `id`s directly; snapshot diff covers the change

## Out of scope

- Q2 `isNational` on `SarOrganisation` — closed on "leave off" default 2026-08-04 pm (see `doc/planning/APG-2546/DELIVERY-LOG.md` "Q2 closed on default"). If Roxanne raises `is_national` later, spin a fresh ticket — do not fold into APG-2546.

## Artefacts

| Item | Value |
|---|---|
| Sample PDF page count | 3 |
| Snapshot files updated | `sar-api-response.json`, `sar-expected-render-result.html` |
| `entity-schema.json` | unchanged (no JPA entity edits) |

## Verification

- `./gradlew ktlintCheck test` — green (678 tests)
- `grep -n "\.id\b" src/main/…/SubjectAccessRequestService.kt` — no remaining `SarPerson.id` / `SarOrganisation.id` reads inside the service (only entity-level reads on `it.id!!` for the batch lookup and mapper receivers, which are correct)
- `grep -n "Person ID\|<td>Id</td>" src/main/resources/sar_template.mustache` — zero hits (the retained `Original referral ID` row inside `{{#originalReferral}}` is scoped to `SarOriginalReferral.id`, not the removed fields)

Working doc: `doc/planning/APG-2546/PR-5-strip-internal-ids.md`.


Alison-Weatherston (approved):
LGTM

